PR #6, opened 2019-10-08
Feat test

mcmcgrath13:


.github/workflows/make-release.yml
```diff
@@ -0,0 +1,73 @@
...
+        npm run build --if-present
+        npm test
+        npm run package:mac
+      env:
+        CI: true
+    - name: add app to release
+      uses: JasonEtco/upload-to-release@master
+      with:
+        args: dist/installers/neuro-task_0.0.1_amd64.deb
+
```
fernandogelin, line 53: is it possible to get the version number from the environment?

PR #27, opened 2020-04-09
feat: add firebase functionality

fernandogelin:


src/App.js
```diff
@@ -18,6 +19,7 @@ if (IS_ELECTRON) {
   psiturk = new PsiTurk(getTurkUniqueId(), '/complete')
   /* eslint-enable */
 }
+const firebase = true;
```
mcmcgrath13: how can this get overwritten to false? should there be an env var?

src/App.js
```diff
@@ -29,6 +31,16 @@ class App extends React.Component {
         <Experiment settings={{
           timeline: tl,
           on_data_update: (data) => {
+            if (firebase) {
+              if (data.trial_index === 1 ) {
```
mcmcgrath13, line 35: isn't there an issue sometimes where we don't have the patient ID on the first trial? and don't get it until a few screens in?

src/App.js
fernandogelin: I think I changed that in the latest commit.. maybe I didn't push?

src/App.js
fernandogelin: first trial is 0. 1 has the patient id. I guess it depends on the task.

src/App.js
fernandogelin: I did push.. maybe the PR didn't update :/

fernandogelin: it says `outdated` here

mcmcgrath13:
I'd make it more abstract with a boolean for `dbInitialized` and then once
we get a trial with a `patient_id` do that and turn it to true

fernandogelin:
Actually no, what needs to change is the function that creates the db document. If it doesn't have a patient id it can still create documents with randomly generated IDs. We don't need to merge this right now, it's mostly to show Wasita what changes need to be made.

src/App.js
```diff
@@ -48,7 +60,8 @@ class App extends React.Component {
               }
               completePsiturk()
             } else {
-              jsPsych.data.get().localSave('csv','neuro-task.csv');
+              
+              // jsPsych.data.get().localSave('csv','neuro-task.csv');
```
mcmcgrath13: why comment this out?

mcmcgrath13 (approved):

PR #34, opened 2020-09-24
feat: 🎸 update readme

Rashi1997:


README.md
```diff
@@ -6,11 +6,11 @@ This repo contains a simple neuro task to use as a starter for building [jspsych
 
 ## Getting Started
 
-1. Clone this repo onto your computer
+1. Clone this repo onto your computer using git
```
mcmcgrath13: should we call out that `project_name` is a placeholder?

mcmcgrath13 (commented):
one minor comment, otherwise looks good!

README.md
Rashi1997: oh yes missed that! I will change that to `task-<TASK NAME>` instead and renaming in next step.

mcmcgrath13 (approved):

PR #36, opened 2020-09-28
Feat task trials

Rashi1997:


Rashi1997:
The tests here are failing because the workflow has `GH_TOKEN` from secrets used for behavioral task trials which is private library, but I guess there is no Secrets defined. I cannot see an option to add a secret.

mcmcgrath13:
@Rashi1997 the behavioral task trial library should be public.  At that point, I think we just need to add the set up step for the github package registry to the workflows.

Rashi1997:
> @Rashi1997 the behavioral task trial library should be public. At that point, I think we just need to add the set up step for the github package registry to the workflows.

@mcmcgrath13 I agree! I have added this to the workflow where the npm install, but does it get the GH_TOKEN from the secrets?
```
- name: npm install
      run: |
        npm config set "//npm.pkg.github.com/:_authToken" ${{ secrets.GH_TOKEN }}
        npm install
```

mcmcgrath13:
> > @Rashi1997 the behavioral task trial library should be public. At that point, I think we just need to add the set up step for the github package registry to the workflows.
> 
> @mcmcgrath13 I agree! I have added this to the workflow where the npm install, but does it get the GH_TOKEN from the secrets?
> 
> ```
> - name: npm install
>       run: |
>         npm config set "//npm.pkg.github.com/:_authToken" ${{ secrets.GH_TOKEN }}
>         npm install
> ```

Do we need that token if it's a pbulic package? (cc @fernandogelin )  Is there any reason it's a custom token instead of the default `GITHUB_TOKEN` one?

Rashi1997:
Works now!🎉

mcmcgrath13:
@Rashi1997 could you resolve the conflicts and make sure this is still working as expected, then I'll merge it in?

Rashi1997:
@mcmcgrath13 @mirestrepo Can you please help me review this before merging?

public/config/version.json
```diff
@@ -0,0 +1 @@
+{"sha":"72480175cdc7411835711287df7d0a48843232a0","ref":"feat-task-trials"}
```
mcmcgrath13: should this be here? I think it might have been written by running the code and we should add it to the .gitignore

package.json
```diff
@@ -54,6 +55,11 @@
     "build:win:clinic": "cross-env REACT_APP_AT_HOME=\"false\" npm run build",
     "build:home": "REACT_APP_AT_HOME=\"true\" npm run build"
   },
+  "config": {
```
mcmcgrath13, line 59: why is this needed?

src/timelines/main.js
```diff
@@ -1,30 +1,29 @@
...
 import taskBlock from './taskBlock'
 
-import { MTURK, lang } from '../config/main'
+import { lang, config } from '../config/main'
 import { practiceBlock } from '../config/practice'
 import { tutorialBlock } from '../config/tutorial'
 import { exptBlock1, exptBlock2 } from '../config/experiment'
 
-
+console.log(config)
```
mcmcgrath13: remove console log

src/timelines/preamble.js
```diff
@@ -1,17 +1,23 @@
...
-import startCode from '../trials/startCode'
-import userId from '../trials/userId'
+import { jsPsych } from 'jspsych-react'
+import { eventCodes, lang } from '../config/main'
 import holdUpMarker from '../trials/holdUpMarker'
-import { AT_HOME, IS_ELECTRON } from '../config/main'
-
+import {showMessage, userId} from '@brown-ccv/behavioral-task-trials'
+import { config, AT_HOME } from '../config/main'
+console.log(showMessage('html_button_response',undefined, config,lang.task.name, false, true, undefined, undefined, [lang.prompt.continue.button]))
```
mcmcgrath13: remove console log

src/timelines/preamble.js
```diff
@@ -1,17 +1,23 @@
...
 console.log('at_home', AT_HOME)
 console.log('env at home', process.env.REACT_APP_AT_HOME)
 const preamble = {
   type: 'html_keyboard_response',
   stimulus: '',
-  timeline: (AT_HOME || !IS_ELECTRON) ?
-    [experimentStart(), userId()] :
-    [experimentStart(), userId(), holdUpMarker(), startCode()]
+  timeline: (!config.USE_PHOTODIODE) ?
+    [showMessage('html_button_response',undefined, config,`<h1>${lang.task.name}</h1>`, false, true, undefined, undefined, [lang.prompt.continue.button]),
```
mcmcgrath13: will show message even not be an `html_button_esponse`? same with userId and `html_keyboard_response` - do those need to be passed in? What are the undefineds?

could this block of code be prettified so it's easier to read?

src/timelines/taskTrial.js
```diff
@@ -1,24 +1,18 @@
...
-    trial_earnings: 0,
-    start_time: Date.now()
-  }
 
   // timeline
   let timeline = [
     // show condition
-    showCondition(condition, 1000, 500),
-    fixation(650),
+    showMessage('html_keyboard_response', 1000, config, condition, true, false, eventCodes.evidence, eventCodes.evidence),
```
mcmcgrath13: why pass the event codes twice?

mcmcgrath13 (changes requested):

mcmcgrath13 (commented):
@Rashi1997 I've been thinking about our set of fields passed here and how to handle it most elegantly, what would you think about taking advantage of object destructuring?

So all of the functions in the library would look something like:
```js
const myFunction = (config, options) => ( ...
```
where the object in the second position will have a default for every item, so then in the function there's a line like
```js
const {id, stimulus, num_blinks} = Object.assign({}, defaults, options) 
```
Where defaults is something like:
```js
const defaults = {
  stimulus = '<h1>Welcome</h1`>,
  num_blinks = 1,
  id = 123
}
```
and then the functions are called like
```js
const stimulus = `<h1>${myVar}</h1>`
const myTimeline = [
  ...,
  myFunction(config, {id: 124, stimulus})
  ...
]
```

I know sometimes there are other required fields, and we could use the positional args for those.

Thoughts?  I know this is probably more feedback than you were hoping for...

Rashi1997:
Thank you @mcmcgrath13 I was just thinking about the same thing, while replacing instances of `undefined, undefined` with `...Array(2)`, so this way I could avoid doing those.

src/config/main.js
```diff
@@ -48,13 +49,15 @@ const defaultBlockSettings = {
 	photodiode_active: false
 }
 
+// setting config for trials
+const config = init({USE_PHOTODIODE: !AT_HOME && IS_ELECTRON,  USE_EEG: false, IS_ELECTRON: IS_ELECTRON, USE_MTURK: MTURK})
```
mcmcgrath13: `IS_ELECTRON: IS_ELECTRON` can be destructured into `IS_ELECTRON`

mcmcgrath13: but should it be USE_ELECTRON in the behavioral task library to be more consistent with the others?

src/timelines/preamble.js
```diff
@@ -1,17 +1,50 @@
...
+  timeline: !config.USE_PHOTODIODE
+    ? [
+        showMessage(config, {
+          responseType: "html_button_response",
+          message: `<h1>${lang.task.name}</h1>`,
+          responseEndsTrial: true,
+          buttons: [lang.prompt.continue.button],
+        }),
+        userId(jsPsych, config, {
+          responseType: "html_keyboard_response",
```
mcmcgrath13: shouldn't `html_keyboard_response` be the default for `userId`? Also, sometimes `message` is a whole stimulus, and sometimes it seems to be just a text string - could you make sure those are consistent on the library side? I'd vote for something like `message` (and similar variants) are just strings, whereas `stimulus` (and variants` are html, and if just message is passed, it's shoved into `h1` tags, but if stimulus is passed, it's ignored - thougths?

src/timelines/taskTrial.js
```diff
@@ -1,33 +1,44 @@
...
-    trial_earnings: 0,
-    start_time: Date.now()
-  }
-
   // timeline
   let timeline = [
     // fixation
-    fixation(650),
+    fixation(config, {
+      responseType: "html_keyboard_response",
```
mcmcgrath13: does fixation need a passed response type? I'd think the default is almost always fine

mcmcgrath13 (changes requested):
the json is definitely a lot cleaner! getting close!

src/config/main.js
Rashi1997: Yes!

src/timelines/preamble.js
Rashi1997: I can definitely revisit this piece of code and use a better way of passing the message

mirestrepo:
You are doing great @Rashi1997 💪😅

src/config/main.js
mcmcgrath13: this one still needs to be resolved

src/timelines/preamble.js
```diff
@@ -1,17 +1,56 @@
...
+import { eventCodes, lang } from "../config/main";
+import holdUpMarker from "../trials/holdUpMarker";
+import { showMessage, userId } from "@brown-ccv/behavioral-task-trials";
+import { config, AT_HOME } from "../config/main";
+console.log("at_home", AT_HOME);
+console.log("env at home", process.env.REACT_APP_AT_HOME);
+console.log(
+  showMessage(config, {
+    responseType: "html_button_response",
+    message: `<h1>${lang.task.name}</h1>`,
```
mcmcgrath13: this message is a full stimulus, but the below messages are just text, should this just be `lang.task.name`?

mcmcgrath13 (changes requested):
so close! one unresolved comment from last time, then as we discussed on slack, make sure every option has a reasonable (if not perfect) default and use those here wherever we can

src/timelines/preamble.js
Rashi1997: Oh missed these! I can remove that console log

mirestrepo:
Is this ready for merging? We should try to close up the open tasks

mcmcgrath13:
> Is this ready for merging? We should try to close up the open tasks

Not yet, @Rashi1997 just merged in the changes to the library, so those changes need be made here too.  Plus adding in the attention tasks.

src/timelines/main.js
```diff
@@ -9,9 +9,9 @@ import { exptBlock1, exptBlock2 } from "../config/experiment";
 
 const primaryTimeline = [
   preamble,
-  countdown({ duration: 1000, text: lang.countdown.message1, time: 3 }),
+  countdown({ duration: 1000, message: lang.countdown.message1, time: 3 }),
```
mcmcgrath13: for all of these uses, remove any keys, where our value matches the default (in the case of countdown, duration and time)

package.json
```diff
@@ -13,10 +13,11 @@
   "homepage": ".",
   "repository": "https://github.com/brown-ccv/honeycomb",
   "dependencies": {
-    "@brown-ccv/behavioral-task-trials": "^1.21.2",
+    "@brown-ccv/behavioral-task-trials": "^1.24.0",
```
mcmcgrath13, line 16: there shouldn't be a `package-lock.json` commited to the repo anymore - can you make sure it's in the gitignore on this branch and delete the one that's commited from being tracked?

package.json
```diff
@@ -13,10 +13,11 @@
...
     "@fortawesome/fontawesome-free": "^5.9.0",
     "bootstrap": "^4.3.1",
     "cross-env": "^6.0.3",
+    "cz-conventional-changelog": "^3.2.0",
```
mcmcgrath13: move to dev dependencies

src/timelines/preamble.js
```diff
@@ -33,11 +33,10 @@ const preamble = {
         userId(jsPsych, config, {
           duration: 800,
```
mcmcgrath13: no duration needed for clinic-based `userId`

src/timelines/preamble.js
```diff
@@ -20,7 +20,7 @@ const preamble = {
...
           setIdMessage: lang.userid.set,
-          defaultPatientId: process.env.REACT_APP_PATIENT_ID,
+          defaultId: process.env.REACT_APP_PATIENT_ID,
```
mcmcgrath13: don't need a `defaultId` for the mturk version

mcmcgrath13 (changes requested):

mcmcgrath13 (approved):
🚀

PR #37, opened 2020-11-17
fix: change master to main in brown-ccv/gh-actions lines in workflows

wasita:
otherwise the gh actions will fail (given what i've seen on task-effort at least and i have since fixed in the same way as i've done in this PR). plz see below:

chore__bump_version_from_1_3_0-beta_to_1_3_1_·_lnccbrown_task-effort_1381a6e


mcmcgrath13:
fixed by #39

PR #68, opened 2021-02-09
Delete beads refs

nicoleprovenza:


src/lib/markup/stimuli.js
```diff
@@ -1,6 +1,11 @@
 const baseStimulus = (element, prompt=false) => {
-  const class_ = (prompt) ? 'main-prompt': 'main'
-  return `<div class=${class_}>${element}</div>`
+  return (
+    `<div class='center_container'>
```
mcmcgrath13, line 3: does the logic on main/main prompt being deleted mess anything else up? `main-prompt` and `center_container` have very similar styling - why not use what was here?

src/lib/utils.js
```diff
@@ -1,5 +1,5 @@
 import { jsPsych } from 'jspsych-react'
-import requireContext from 'require-context.macro'
+//import requireContext from 'require-context.macro'
```
mcmcgrath13: why comment this out? wouldn't leaving it here but with no images to find just  return an empty array?

mcmcgrath13 (changes requested):

src/lib/markup/stimuli.js
nicoleprovenza: Main prompt was putting the stimulus at the top of the screen rather than the center. I wasn't sure how else to center the stimulus (the letters)

src/lib/utils.js
nicoleprovenza: A bug came up when there were no images. I got stuck on this last night... I think I am going to put 2 blank images in the assets folder (instead of the moneybags) and revert all the changes I made surrounding images

src/lib/markup/stimuli.js
mcmcgrath13: is main-prompt used anywhere else? should it be deleted now if you aren't using it here?

nicoleprovenza: no, it's not used anywhere else. (Other than where it's defined in the app.css file). I don't have a strong opinion

mcmcgrath13: that's fine, but can you go through the css file and make sure that we're using what's in there/ delete anything that's unused?

nicoleprovenza:
closing the branch because Rashi fixed this in her latest PR.

PR #69, opened 2021-02-09
ci: refactor build and installers

mirestrepo:
- On PR and push to main run only the build by default
- On dispatch we have the options of building installers for clinic-win, home-win, clinic-all, home-all
- On release, clinic-all and home-all


mirestrepo:
@mcmcgrath13 `package.json` has different scripts for `npm run build:win:home` and others like that, but in the original workflows only `npm run build` was used... Should either work? Does the environment variable in CI propagate as needed? I see that for windows in the `package.json` it add a `cross-env`.

The automated builds are passing, but well'll have to merge before we can test the dispatch, and on -release/tag ones

mirestrepo:
superseeded by  #70

PR #78, opened 2021-06-10
feat: port env variables and github actions

suivo2001:


.github/workflows/release.yml
```diff
@@ -20,77 +17,79 @@ jobs:
...
-      if: startsWith(matrix.os, 'mac')
-      uses: svenstaro/upload-release-action@v2
-      with:
-        repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: dist/installers/${{ steps.package_info.outputs.package_name }}-${{ steps.package_info.outputs.package_version }}.dmg
-        asset_name: ${{ steps.package_info.outputs.package_name }}-${{ steps.package_info.outputs.package_version }}-${{ matrix.setting }}.dmg
-        tag: ${{ github.ref }}
+      - name: Upload app to release - windows
+        uses: svenstaro/upload-release-action@v2
+        with:
```
mcmcgrath13, line 69: all of the `if` statements seem to have been lost with this update, was that on purpose?

public/electron.js
```diff
@@ -163,21 +157,28 @@ ipc.on('trigger', (event, args) => {
...
+ipc.on('syncCredentials', (event) => {
+  event.returnValue = {envParticipantId: process.env.REACT_APP_PARTICIPANT_ID, envStudyId: process.env.REACT_APP_STUDY_ID}
+})
+
 // listener for new data
 ipc.on('data', (event, args) => {
 
   // initialize file - we got a patinet_id to save the data to
-  if (args.patient_id && fileName === '') {
+  if (args.participant_id && args.study_id && fileName === '') {
```
mcmcgrath13, line 177: typo/switch to participant on line 176

src/timelines/preamble.js
```diff
@@ -1,46 +1,27 @@
...
+import startCode from "../trials/startCode";
+import welcome from "../trials/welcome";
+import adjustVolume from "../trials/adjustVolume";
+import camera from "../trials/camera";
+import { USE_EVENT_MARKER, VOLUME, VIDEO } from "../config/main";
+
+let timeline = [];
+if (VOLUME) {
+  timeline.push(adjustVolume());
+}
```
mcmcgrath13: this timeline doesn't look like what's in the old part of the diff

mcmcgrath13 (changes requested):
For some background, the `config` object here is to play nicely with the behavioral task trial library.  We should lean on that library where it makes sense (maybe but the video start/stop there) and also continue to use that config setup for consistency.  It might make sense to add the other features to that config so they can be leveraged in the library as well.

src/timelines/preamble.js
suivo2001: @mcmcgrath13 About the config object, should I just use the variables from the object when they are needed, like instead of using USE_PHOTODIODE directly, I import the config object and use config.USE_PHOTODIODE instead? Also what variables should I add? Maybe USE_CAMERA?

mcmcgrath13: yup, definitely USE_CAMERA.  Any others that are used in trials, but in the feature category (env vars).

suivo2001: @mcmcgrath13 Thanks. Also, relating back to the original comment, what do we want the preamble to look like. Right now I just followed what was in MSIT, is there a specific order we want the screens?

mcmcgrath13: Just recreate the preamble that was already in honeycomb

public/electron.js
```diff
@@ -163,21 +157,28 @@ ipc.on('trigger', (event, args) => {
...
-  if (args.patient_id && fileName === '') {
+  // initialize file - we got a participant_id to save the data to
+  if (args.participant_id && args.study_id && fileName === '') {
     const dir = app.getPath('userData')
-    patientID = args.patient_id
-    fileName = `pid_${patientID}_${Date.now()}.json`
+    participantID = args.participant_id
+    studyID = args.study_id
+    fileName = `pid_${participantID}_${Date.now()}.json`
     filePath = path.resolve(dir, fileName)
```
mcmcgrath13, line 181: `studyId` isn't used anywhere - should it be?  is the `filePath` resolution correct?

public/electron.js
```diff
@@ -195,11 +196,30 @@ ipc.on('data', (event, args) => {
...
     if (args.trial_type === 'image_keyboard_response') images.push(args.stimulus.slice(7))
   }
 })
 
+// Save Video
+ipc.on('save_video', (event, fileName, buffer) => {
+  
+  const desktop = app.getPath('desktop')
+  const name = app.getName()
+  const today = new Date(Date.now())
```
mcmcgrath13: `Date.now()` not needed, `new Date()` returns the now value

public/electron.js
```diff
@@ -195,11 +196,30 @@ ipc.on('data', (event, args) => {
...
+  const date = today.toISOString().slice(0,10)
+  const fullPath = path.join(desktop, dataDir, studyID, participantID, date, name, fileName)
```
mcmcgrath13, line 210: this path doesn't match the general data saving path, but they probably should.  This one is more specific so maybe go with this one?

mcmcgrath13: or maybe the one below is the correct place and this can be ignored

src/App.js
```diff
@@ -1,73 +1,191 @@
...
-import { tl } from './timelines/main'
-import { config } from './config/main'
-import './App.css'
-import 'bootstrap/dist/css/bootstrap.css'
-import '@fortawesome/fontawesome-free/css/all.css'
-import { getTurkUniqueId, sleep } from './lib/utils'
-import { addToFirebase, createFirebaseDocument } from './firebase.js'
+import React, { useState, useEffect, useCallback } from "react";
+import "./App.css";
+import "bootstrap/dist/css/bootstrap.css";
```
mcmcgrath13: file should be `App.jsx`

src/App.js
```diff
@@ -1,73 +1,191 @@
...
+import "@fortawesome/fontawesome-free/css/all.css";
+import Login from "./components/Login";
+import { jsPsych } from "jspsych-react";
```
mcmcgrath13: group the imports by type (external libraries, then css, then components, then internal functions)

src/components/JsPsychExperiment.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Experiment } from "jspsych-react";
```
mcmcgrath13: file extension on this and login should be `.jsx` (is that what we did on the tasks?)

src/config/main.js
```diff
@@ -2,62 +2,88 @@
...
+  repeats_per_condition: 1, // number of times every condition is repeated
+  is_practice: false,
+  is_tutorial: false,
+  photodiode_active: false,
+};
 
 // setting config for trials
-const config = init({USE_PHOTODIODE: !AT_HOME && IS_ELECTRON,  USE_EEG: false, USE_ELECTRON: IS_ELECTRON, USE_MTURK: MTURK})
+const config = init({
+  USE_PHOTODIODE: USE_PHOTODIODE,
```
mcmcgrath13: the variables that are the same name can use the `{ USE_PHOTODIODE, ... }` syntax

mcmcgrath13: which might just be `USE_PHOTODIODE` at this point... should the original declaration of these vars match the config names so that this mapping is cleaner?

src/trials/camera.js
```diff
@@ -0,0 +1,112 @@
...
+      })
+    
+    },
+    on_finish: () => {
+      if (config.USE_CAMERA) {
+        try {
+          window.cameraCapture.start()
+          window.screenCapture.start()
+        } catch (error) {
+          window.alert("Camera permissions were not given, if you choose to proceed, your recording will not be saved. Please restart the experiment after you have given permission.")
```
mcmcgrath13, line 104: instead of have the camera stopping as part of `experimentEnd` trial, create a mini trial in here that does that and add it to the timeline.  Then restore the original trial start/stop `showMessage` from current honeycomb and delete the `experimentStart` and `experimentStop` trials.

src/timelines/main.js
```diff
@@ -15,6 +16,7 @@ import {
 } from "../trials/quizTrials";
 
 const primaryTimeline = [
+  experimentStart(),
   preamble,
```
mcmcgrath13, line 19: add the `camera` trial to the timeline if video is enabled (should that check be within the `camera.js` file or here do you think?

mcmcgrath13 (changes requested):

public/electron.js
suivo2001: Yeah it should be in the path but not in the name itself

public/electron.js
suivo2001: The path for the video and the data seems the same to me.

src/timelines/main.js
suivo2001: I don't see how it would be done in camera.js. The main timeline seems good to me.

mcmcgrath13 (approved):
:shipit:

mcmcgrath13:
@suivo2001 bump the version in package.json when you're ready to merge

PR #80, opened 2021-06-21
fix: fix electron env not loading in packaged build

suivo2001:
In the packaged distribution version, the pop-up for the event marker not connected doesn't show up in the clinic version. This was due to environment variables not loading in `electron.js`. This is a fix by using a json object instead of env variables and writing the json file with a npm script.


mcmcgrath13 (commented):
Instead of using another config file, maybe instead we should have the react app send it's electron related config (use event marker - maybe others) back to the electron process at app startup and have a listener in electron.js for those.  That guarantees they're in sync without another file/script.

mirestrepo:
> Instead of using another config file, maybe instead we should have the react app send it's electron related config (use event marker - maybe others) back to the electron process at app startup and have a listener in electron.js for those. That guarantees they're in sync without another file/script.

I don't mind that idea. From my point of view may main concern is to not allow/require variables that should match at build/run time to be (or needed to be) set/modified at run-time.

mcmcgrath13:
> I don't mind that idea. From my point of view may main concern is to not allow/require variables that should match at build/run time to be (or needed to be) set/modified at run-time.

Agreed.  I think this change would make all environment variables either build time or runtime and not both, which would make the docs/usage a little clearer.

PR #81, opened 2021-06-22
feat: moved event marker to buildtime

suivo2001:
Removed `REACT_APP_USE_EVENT_MARKER `from `electron.js`. This variable should be buildtime only now. I put the popup check in a new page `checkEventMarker.js`. It seems like the App component loads twice at start. I read that conditional rendering could contribute to that. In the other cases in the app this doesn't matter, but if we put the popup there, it pops up twice. I could move it to right after the login screen if that's better.


mcmcgrath13:
@suivo2001 I'm not sure I understand the motivation behind this PR.  As the event marker checks/forcing quit/retrying is all on the electron.js side.  Could you talk more about what the broader goal is here?

suivo2001:
@mcmcgrath13 As honeycomb stands now, and the other tasks, REACT_APP_USE_EVENT_MARKER has to be set as a runtime env variable and a build time one. It seems like the env variables checked in `electron.js` is not packaged at build time. The issue was that the event marker popup doesn't show up because of that unless we explicitly set the env variable at runtime. It's not really a problem but I think we discussed that REACT_APP_USE_EVENT_MARKER doesn't need to be a runtime env variable, so I moved the check for popup into react instead of staying in `electron.js`, so it is now only a build time env variable. I actually realized that I could make the popup show once at startup by passing an empty array for the `UseEffect` in App, so I could push that as well.

mcmcgrath13:
Instead of moving the `REACT_APP_USE_EVENT_MARKER` check to the frontend, I think it would be better to send the build time value of that env var (and others that should be build time only, but are used in electron) back to the electron.js file.  So the variables in electron.js would default to `false` then when the backend process receives the build time values from the front end, would update them and use those values.

suivo2001:
I just tried that with a new environment.js file that exports env variables, it behaves the same as calling the variable directly from electron.

suivo2001:
@mcmcgrath13 the changes are ready to be reviewed! I changed everything to USE_EEG, including the env variable, and added a check for video under save_video as well.

src/App.jsx
```diff
@@ -88,11 +88,11 @@ function App() {
     console.log("Photodiode:", config.USE_PHOTODIODE);
     // If on desktop
     if (config.USE_ELECTRON) {
-      const electron = window.require("electron");
-      const renderer = electron.ipcRenderer;
-      setRenderer(renderer);
+      const ipcRenderer = window.require("electron").ipcRenderer;
```
mcmcgrath13: could we be fancy and do `const { ipcRenderer} = window.require("electron")` ? (literally not sure if that works, but it seems like it might)

src/App.jsx
```diff
@@ -141,7 +141,8 @@ function App() {
         setReject(true);
       }
     }
-  }, [setLoggedIn, query]);
+  // eslint-disable-next-line 
+  }, []);
```
mcmcgrath13, line 145: why remove the dependencies? were they not needed? or is it that we just want this to run once only no matter what?

src/lib/markup/photodiode.js
```diff
@@ -39,11 +39,11 @@ const pdSpotEncode = (taskCode) => {
     }
 
 		if (config.USE_PHOTODIODE) {
-				const blinkTime = 40
-				let numBlinks = taskCode
-		    if (taskCode < eventCodes.open_task) numBlinks = 1;
-				repeat_pulse_for(blinkTime, numBlinks)
-				if ( ipcRenderer ) ipcRenderer.send('trigger', taskCode)
+      const blinkTime = 40
```
mcmcgrath13: indentation looks funny here

src/lib/markup/photodiode.js
```diff
@@ -39,11 +39,11 @@ const pdSpotEncode = (taskCode) => {
...
+      let numBlinks = taskCode
+      if (taskCode < eventCodes.open_task) numBlinks = 1;
+      repeat_pulse_for(blinkTime, numBlinks)
+      if ( ipcRenderer ) ipcRenderer.send('trigger', { taskCode })
```
mcmcgrath13: is the on `trigger` set up for this change? why make this an object?

suivo2001: Yes I changed that. I used to pass in USE_EVENT_MARKER here , but I didn't change it back.

src/App.jsx
suivo2001: The dependencies made the app render twice at startup, and the pop up shows up twice. The dependencies should not be needed since it's always on the `App` page anyways. I tested with the query in firebase and the env participant ID and study ID and it works the same.

suivo2001:
@mcmcgrath13 should I change this version to 2.0.1?

src/App.jsx
suivo2001: yes it works

mcmcgrath13:
> @mcmcgrath13 should I change this version to 2.0.1?

I'd call this 2.1.0 since I'd say the build time only variables is a new feature

mcmcgrath13 (approved):
:shipit:

PR #82, opened 2021-06-29
feat: studyID as directory and fix video bug

suivo2001:


public/electron.js
```diff
@@ -217,11 +225,7 @@ ipc.on('data', (event, args) => {
 // Save Video
 ipc.on('save_video', (event, fileName, buffer) => {
   if (VIDEO){
-    const desktop = app.getPath('desktop')
```
mcmcgrath13, line 220: this is a nit I didn't really care about on the task, but for honeycomb I think we should fix.  We have a global variable `fileName`, but this handler shadows it.  Can we change the variable in the `save_video` function to be `videoFileName` or similar?

mcmcgrath13 (changes requested):
one nit, otherwise looks good!

mcmcgrath13 (approved):
🐝

PR #84, opened 2021-07-15
refactor: preventing error when online and firebase not enabled

galenwinsor:


src/App.jsx
```diff
@@ -72,127 +74,109 @@ function App() {
...
+      } else if (config.USE_PROLIFIC) {
+        const pID = getProlificId()
         if (config.USE_FIREBASE && pID) {
-          setMethod("firebase");
-          setLoggedIn(true, "prolific", pID);
+          setMethod('firebase')
+          setLoggedIn(true, 'prolific', pID)
         } else {
-          setReject(true);
+          setMethod('online')
```
mcmcgrath13: I think this one should still be reject, local save + prolific doesn't make sense (no way to get the data from the participant)

mcmcgrath13 (changes requested):

mcmcgrath13 (approved):

PR #87, opened 2021-07-22
refactor: cleaning up path logic in electron.js

galenwinsor:
Just because I had a little extra time!


public/electron.js
```diff
@@ -225,7 +242,7 @@ ipc.on('data', (event, args) => {
 // Save Video
 ipc.on('save_video', (event, videoFileName, buffer) => {
   if (VIDEO){
```
mcmcgrath13, line 250: should this check for the save path? or is it guaranteed to exist already?

mcmcgrath13 (approved):
🗃️

PR #89, opened 2021-07-27
Feat stroop

galenwinsor:
Stroooppppp

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -42,7 +42,7 @@ html, body, #root, .App {
   width: 100vw;
 }
 
-.width-min-content {
+.width-max-content {
   width: max-content;
 }
 
--- a/src/App.css
+++ b/src/App.css
@@ -18,6 +18,10 @@ html, body, #root, .App {
   height: 100%;
 }
 
+.margin-centered {
+  margin: 0 auto;
+}
+
 .height-100 {
   height: 100%;
 }
@@ -38,8 +42,24 @@ html, body, #root, .App {
   width: 100vw;
 }
 
+.width-min-content {
+  width: max-content;
+}
+
 .font-size-large {
+  font-size: 2.5rem;
+}
+
+.font-size-extra-large {
   font-size: 4rem;
+}
+
+.red-font {
+  color: red;
+}
+
+.blue-font {
+  color: blue;
 }
 
 .photodiode-box {
--- a/src/App.css
+++ b/src/App.css
@@ -18,12 +18,28 @@ html, body, #root, .App {
   height: 100%;
 }
 
+.height-100 {
+  height: 100%;
+}
+
+.height-100-view {
+  height: 100vh;
+}
+
 .width-50 {
   width: 50%;
 }
 
 .width-100 {
   width: 100%;
+}
+
+.width-100-view {
+  width: 100vw;
+}
+
+.font-size-large {
+  font-size: 4rem;
 }
 
 .photodiode-box {
--- a/src/config/experiment.js
+++ b/src/config/experiment.js
@@ -21,9 +21,9 @@ const getLocalConfig = async (participantID, studyID) => {
     }
     renderer.send("save-config", experimentConfig, participantID, studyID)
   } else if (envConfig.USE_FIREBASE) {
-    const newConfig = await firestoreConfig(studyID, participantID);
+    const newConfig = await firestoreConfig(studyID, participantID)
     if (newConfig) {
-      experimentConfig = newConfig;
+      experimentConfig = newConfig
     }
   } else {
     console.warn("Using default config")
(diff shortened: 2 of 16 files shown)


src/config/experiment.js
```diff
@@ -30,35 +29,39 @@ const getConfig = async (participantID, studyID) => {
...
  */
-const overrideSettings = (override, defaultBlockSettings) => {
-  console.log("Default block settings:", defaultBlockSettings)
-  console.log("override:", override)
-  const newBlock = deepCopy(defaultBlockSettings)
-  Object.assign(newBlock, override)
-
-  return newBlock
-}
+// const overrideSettings = (override, defaultBlockSettings) => {
```
mcmcgrath13, line 50: why aren't they used? When would they be used? provide a bit more guidance here

src/trials/choice.js
```diff
@@ -0,0 +1,31 @@
...
+    // A type of trial that calls a function rather than display a stimulus.
+    type: "html_keyboard_response",
+    // Allows the trial to be manually ended with the done() function.
+    response_ends_trial: true,
+    stimulus:
+      `<div class="width-100-view height-100-view">
+            <p id="color-display" class="centered-h-v font-weight-bold font-size-large" style="color:${color}">${word}</p>
+        </div>`,
+    on_start: () => {
+      removeCursor("experiment")
```
mcmcgrath13, line 25: does the cursor need to be added back?

src/trials/showEarnings.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Displays the earnings from a trial.
+ * @param {number} duration How long to display the earnings.
+ * @returns trial The jsPsych trial object.
+ */
+const showEarnings = (duration) => {
+  // Maps colors to their corresponding key code.
+  const colorKeyMappings = {
+    red: 82,
```
mcmcgrath13: should the key map be part of the config instead? if someone added `'green'` to the config, would it work?

src/trials/choice.js
galenwinsor: It is in `showEarnings`

src/trials/showEarnings.js
galenwinsor: Actually fixed this in the next commit!

src/App.css
```diff
@@ -26,6 +38,30 @@ html, body, #root, .App {
   width: 100%;
 }
 
+.width-100-view {
+  width: 100vw;
+}
+
+.width-min-content {
+  width: max-content;
```
mcmcgrath13, line 46: class and css don't really match

src/timelines/preamble.js
```diff
@@ -1,25 +1,54 @@
...
-if (envConfig.USE_PHOTODIODE) {
-  timeline.push(holdUpMarker())
-  timeline.push(startCode())
-}
+const preamble = (experimentConfig) => {
+  const colors = experimentConfig.conditions
+  let colorList = ""
+
+  for (let i = 0; i < colors.length; i++) {
+    colorList += `<li class="font-size-large"><span style="color: ${colors[i]}">WORD</span>: ${colors[i][0].toUpperCase()}</li>`
```
mcmcgrath13: isn't the color list a flat array, how does the taking the `[0]` index work?

mcmcgrath13: oh, I see - it's the first key - add a comment

src/timelines/taskBlock.js
```diff
@@ -5,14 +5,8 @@ const taskBlock = (blockSettings) => {
   // initialize block
 	const startingOpts = generateStartingOpts(blockSettings)
 
-  const blockDetails = {
-	  block_earnings: 0.0,
-		optimal_earnings: 0.0,
-		continue_block: true
-	}
-
 	// timeline = loop through trials
```
mcmcgrath13, line 8: why not collect the overall earnings anymore?

src/trials/choice.js
mcmcgrath13: add a comment on that

src/trials/showEarnings.js
```diff
@@ -0,0 +1,47 @@
...
+      } else if (keyPressed === lastColor[0].toUpperCase()) {
+        trial.stimulus = earningsDisplay(1)
+      } else {
+        trial.stimulus = earningsDisplay(-1)
+      }
+      if (envConfig.USE_PHOTODIODE) trial.stimulus += photodiodeGhostBox();
+    },
+    on_finish: (data) => {
+      data.code = eventCodes.show_earnings
+      addCursor("experiment");
```
mcmcgrath13: add a comment on where this was removed

src/App.css
galenwinsor: Yup, oops!

src/timelines/taskBlock.js
galenwinsor: I don't think it was ever used anywhere.

src/trials/choice.js
mcmcgrath13: instead of commenting on what's happening, comment on when we add the cursor back

mcmcgrath13 (changes requested):

mcmcgrath13 (approved):
🔴 🟦 💛

PR #88, opened 2021-07-26
Refactor participant specific config

galenwinsor:

PR #90, opened 2021-07-29
Chore documentation

galenwinsor:


src/trials/welcome.js
```diff
@@ -2,7 +2,7 @@ import { lang } from '../config/main'
 import { photodiodeGhostBox } from '../lib/markup/photodiode'
 import { baseStimulus } from '../lib/markup/stimuli'
 
-const pleaseBiggen = () => {
+const pleaseBegin = () => {
```
mcmcgrath13: this one isn't a typo (though perhaps not well named either) - it's asking the user to make the window larger

mcmcgrath13 (changes requested):

src/trials/welcome.js
galenwinsor: Haha, I see. Fixed!

mcmcgrath13 (approved):
🆎

RobertGemmaJr:
@digicosmos86 @mirestrepo Do you think we should merge this in as well? I think the documentation would be really nice but to have but I don't know enough to know if it's out of date at this point

Actually it's base is #88 so I say we merge it there and take a deeper look in that branch?

mirestrepo:
> @digicosmos86 @mirestrepo Do you think we should merge this in as well? I think the documentation would be really nice but to have but I don't know enough to know if it's out of date at this point
> 
> Actually it's base is #88 so I say we merge it there and take a deeper look in that branch?

Sounds like a plan!

PR #97, opened 2022-06-27
React scripts 5

benjamin-heasly:
Hello!  This PR is in preparation for upgrading to jspsych 7 (which will be my next PR).  This one upgrades Honeycomb to use react-scripts v5 (we were on v3), which in turn gives us webpack v5.

I didn't set out to upgrade webpack! :stuck_out_tongue:  But after resisting it / being in denial for a while, I do think we require v5.

The reason is that jspsych and its plugins come in ES modules that have non-default entry points -- i.e., the module code comes from a file like `node_modules/jspsych/dist/index.js` rather than the default like `node_modules/jspsych/index.js` ([source](https://github.com/jspsych/jsPsych/blob/main/packages/jspsych/package.json#L7)).  Webpack didn't support non-default entry points like this until v5.

We could work around the entry point in our own code by importing `from 'jspsych/dist'`.  But the jspsych plugin modules also need to import jspsych, and they don't do it this way.  They directly import `from 'jspsych'` and seem to rely on the tooling like webpack to find the configured entry point for them ([source](https://github.com/jspsych/jsPsych/blob/main/packages/plugin-html-keyboard-response/src/index.ts#L1)).  So, here we go!

Webpack 5 introduces some upgrades and breaking changes, so this PR also includes various cleanup to get the build working and clean again.

The big breaking change is around polyfills for client code that tries to access modules intended for the server, for example `dotenv` ([cries of pain](https://github.com/facebook/create-react-app/issues/11756)).  We had one instance of client code trying to use `dotenv`, which in turn requires `fs`, which is no longer present for the client.  I am hoping we don't actually need to do this, and can instead rely on env vars that react-scripts exposes to us, the ones that get prefixed with `REACT_APP_`.

I'll put a few more comments in the actual change sets.

:smile_cat:


.eslintrc.js
```diff
@@ -21,5 +22,6 @@ module.exports = {
     'react'
   ],
   'rules': {
+    'react/prop-types': 'off'
```
benjamin-heasly, line 25: Once eslint started running again, I got a bunch of prop-types errors.  IMO the solution with [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html) is kinda clunky.  Since we don't have a ton of React code, I just disabled the check.  But I'm happy to revisit this.

package.json
```diff
@@ -15,12 +15,12 @@
   "dependencies": {
     "@brown-ccv/behavioral-task-trials": "^1.0.0",
     "@fortawesome/fontawesome-free": "^5.9.0",
-    "bootstrap": "^4.3.1",
+    "bootstrap": "^5.2.0-beta1",
```
benjamin-heasly, line 18: It's awkward to depend on a beta, but I wanted to see a clean build!  I am hoping we can just upgrade to 5.2.0 when it becomes GA / non-beta.

package.json
```diff
@@ -71,9 +71,6 @@
       "path": "./node_modules/cz-conventional-changelog"
     }
   },
-  "eslintConfig": {
-    "extends": "react-app"
-  },
```
benjamin-heasly, line 76: I think this section was redundant with `.eslintrc.js`, and it confused me for a while, so I removed it.

package.json
```diff
@@ -98,17 +95,11 @@
     "electron-devtools-installer": "^3.2.0",
     "electron-packager": "^14.0.6",
     "electron-rebuild": "^3.2.7",
-    "eslint": "^5.16.0",
-    "eslint-plugin-import": "^2.18.2",
-    "eslint-plugin-node": "^9.2.0",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-react": "^7.14.3",
-    "eslint-plugin-standard": "^4.0.1",
```
benjamin-heasly, line 106: These conflicted with dependencies from react-scripts v5.  Could we just accept the config and conventions from react-scripts?

package.json
```diff
@@ -98,17 +95,11 @@
...
     "firebase-tools": "^11.0.1",
     "require-context.macro": "^1.1.1"
   },
   "babel": {
-    "plugin": [
+    "plugins": [
```
benjamin-heasly, line 102: This typo spammed the build with lots of confusing errors about all of our js files having an error on line `0`!  Once I added the `s`, the linter started yelling at me about actual code/style issues, so I think that was progress :smile_cat:

src/config/main.js
```diff
@@ -7,6 +7,7 @@ import _ from "lodash";
 import { eventCodes } from "./trigger";
 import { init } from "@brown-ccv/behavioral-task-trials";
 import { getProlificId } from "../lib/utils";
+import app from './app-version'
```
benjamin-heasly: Grab app name and version from the `src/config/app-version.json`, which is created as a postinstall script, and gets its values out of `package.json`.

package.json
```diff
@@ -15,12 +15,12 @@
...
     "@brown-ccv/behavioral-task-trials": "^1.0.0",
     "@fortawesome/fontawesome-free": "^5.9.0",
-    "bootstrap": "^4.3.1",
+    "bootstrap": "^5.2.0-beta1",
     "electron-log": "^3.0.9",
     "electron-squirrel-startup": "^1.0.0",
     "event-marker": "git+https://github.com/brown-ccv/event-marker.git",
     "execa": "^5.0.0",
-    "firebase": "^8.2.6",
+    "firebase": "^9.8.4",
```
benjamin-heasly, line 23: Firebase v9 imported cleaner than v8 ([known issue](https://github.com/firebase/firebase-js-sdk/issues/3711)), so I just went ahead and upgraded the dependency and used "compat" imports to work with our existing code.

src/firebase.js
```diff
@@ -1,7 +1,6 @@
-import firebase from "firebase";
-import 'firebase/firestore';
+import firebase from 'firebase/compat/app';
+import 'firebase/compat/firestore';
 
-require("dotenv").config();
```
benjamin-heasly, line 4: With webpack v5, we are no longer able to access dotenv (and in turn fs) on the client side.  I am hoping we don't need to do this, and can instead rely on react-scripts to provide us with vars (the ones that start with REACT_APP_).

.eslintrc.js
mirestrepo: I think this okay

mirestrepo (commented):
This is great Ben! I’m on vacation until July 18 so I’m going to ask @hollandjg and @eldu to help providing a reviw

src/firebase.js
```diff
@@ -29,17 +28,18 @@ if (window.location.hostname === "localhost") {
...
     .collection('data')
     .doc(startDate)
-    .set({ start_time: startDate, app_version: window.navigator.appVersion, app_platform: window.navigator.platform, results: []})
-    .then(()=>{
+    .set({ start_time: startDate, app_version: window.navigator.appVersion, app_platform: window.navigator.platform, results: [] })
+    .then(() => {
       return true
     })
     .catch((error) => {
+      console.log(error)
```
eldu: You could print these errors with `console.error`

```suggestion
      console.error(error)
```

src/config/main.js
eldu: If the version in packages.json gets updated, how does the `app-version.json` stay in sync?

eldu (commented):
👋 hello! I took a look at this pr, I put in a few comments. Overall the code looks good! 

I'll approve after my question for keeping the app version in sync between the package.json and the app-version.json is answered

src/firebase.js
benjamin-heasly: Oh nice one, thanks.  I'll change it.

src/config/main.js
benjamin-heasly: Hi @eldu :wave: Yeah, I thought about this too and I'd be open to taking a different approach.

My thinking with this approach is that the `app-version.json` is in `.gitignore` so for a fresh checkout it won't exist until it's created by `npm install`, and at that moment it will be in sync.  I think this covers the case of builds created by automation, like GitHub Actions, so at least that important case should work out OK.

I agree it's possible for these to get out of sync during local development, though, and there's nothing at the moment to prevent that.  I think it's a low-tech, best-effort approach similar to how Honeycomb is already handling Git version info in `git-version.js` (called `version.js` before this PR).  In both cases, I think automation should be OK, but local dev could get out of sync.

Do you think we should revisit this?  I was nervous about importing `package.json` directly as a whole, but maybe this is not as bad as I thought?

eldu: Importing package.json doesn't seem too bad to me especially since this repo is public anyway? I'm asking around in CCV as well and we'll see what others think! 

Could also just import what you needed from the `package.json`
`import { name, version } from "package.json"`;

benjamin-heasly: OK, cool, thanks.

Thinking back, the only reason I changed this part of the code is that after the upgrade I started getting warnings like this:

> Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)

And then trying to follow up on that, it seemed like we might not want to import the whole thing, as in this [so](https://stackoverflow.com/questions/64993118/error-should-not-import-the-named-export-version-imported-as-version).

So I thought maybe another approach would be good.  But I'm also happy to revert this part and maybe open a separate issue to keep an eye on that warning?

hollandjg: I'm wondering how this version parameter is actually intended to be used, as it seems a pretty complicated way to get the version number out. Is there a problem if it gets out of sync locally? perhaps storing alongside data collected from a study, to show which version of the code was used when collecting the data. Or is it not intended for that purpose? 

Is there something like this [so post](https://stackoverflow.com/a/22339262/5542581) which might be fit for purpose?

package.json
hollandjg: It looks like `eslint` has been removed completely as a dev dependency, but there's still an `.eslintrc`. Might upgrading to the newest `eslint` and plugins help?

src/config/main.js
benjamin-heasly: Hi @hollandjg.  Yeah, I'm not sure how the parameter is used either, but it does seem to be stored along with task data, so that would be my guess as well.

I think the env vars approach might work.  If I understand it, we'd add some lines to each of our .env files like these:
```
REACT_APP_VERSION=$npm_package_version
REACT_APP_NAME=$npm_package_name
```
And this might let us access the values via `process.env`.  But I don't know if the `npm_` variables would be set when running as an Electron app, so we might need another approach in that case, so I'm not sure if this would simplify things overall.

For now, I think I will back this change out and go with ~named imports~ a default import from package.json, like @eldu suggested and similar to what Honeycomb does already.  I'll open a separate ticket to capture the fact that with this upgrade, Webpack is starting to ~warn~ error about the named import approach, and I'll add some of what we discussed here.

This might be worth tackling on its own, and I didn't really mean to make waves with this part of the PR! :smile:

benjamin-heasly: I backed out the changes with the postinstall script and went with an import from package.json.  If it's OK with you all, maybe we could decouple the app version detection from the upgrades to react-scripts and webpack?  I created https://github.com/brown-ccv/honeycomb/issues/99 to keep track of the app version detection discussion.

package.json
benjamin-heasly: We do still have eslint and plugins as dev dependencies, but we're getting them as the opinionated defaults from react-scripts instead of choosing our own.  So the `.eslintrc.js` is still being used, and I think it's working OK now.

src/config/main.js
eldu: Sounds good!

eldu (approved):
This looks good! Thanks for the updates! And thanks for making the separate issue for getting the version

benjamin-heasly:
Awesome, thanks!  I'll go ahead and merge this and put up my next PR with the actual jspsych 7 stuff.

PR #100, opened 2022-07-13
Electron upgrade

### LINKED ISSUE #92: Upgrade electron
We're currently on a pretty old version of electron, it would be good to get this to the latest stable version.

We were previously held back by a not-yet-done rewrite of electron forge, which we use to do our various builds.  That should be done by now, so we can upgrade electron and electron forge.

All of the electron code is housed in `public/electron.js`, so if there are any breaking api changes, it should be pretty isolated to that file.

benjamin-heasly:
Greetings!  Here's a PR to upgrade honeycomb to use the latest electron and electron-forge tooling.  This is intended to close #92 .

I ran through the electron-forge [Getting Started](https://www.electronforge.io/) docs to learn what are the official recommended dependencies and config for electron these days.  Then I tried to map these back to the honeycomb project and replace our older dependencies with those official recommendations.

All the existing honeycomb build scripts should still run.  The only scripts that needed major changes were for the actual package/postpackage.  The new style seems to be simpler scripts, and a bunch of new "forge" config in package.json.

Running locally on Linux, I was able to `package:linux` and get a deb package created and installed.  I was able to run it and play through the demo task!

Using GitHub actions, I was also able to create and "release" packages for mac and win ([on my own hacked-up honeycomb repo](https://github.com/benjamin-heasly/hello-honeycomb/releases/tag/test12)).  I tried the dmg on my iMac and it also worked!

I don't have a Windows machine to test with, so I'm less sure about that one.

The new electron-forge "Makers" create output files with slightly different names than before, and there seems not to be a way to specify the output folder or names.  So I updated the GitHub actions to match the new names.  I ran into one slight issue with hyphens vs underscores in the created file names on win32, so I created an issue over at [gh-actions](https://github.com/brown-ccv/gh-actions/issues/39).

### DIFF
--- a/public/electron.js
+++ b/public/electron.js
@@ -2,7 +2,6 @@
 if(require('electron-squirrel-startup')) return 
 
 // Modules to control application life and create native browser window
-const imageKeyboardResponse = require('@jspsych/plugin-html-keyboard-response')
 const { app, BrowserWindow, dialog } = require('electron')
 const path = require('path')
 const ipc = require('electron').ipcMain
@@ -238,7 +237,7 @@ ipc.on('data', (event, args) => {
     stream.write(JSON.stringify({...args, git}))
 
     // Copy provocation images to participant's data folder
-    if (args.trial_type === imageKeyboardResponse) images.push(args.stimulus.slice(7))
+    if (args.trial_type === "image-keyboard-response") images.push(args.stimulus.slice(7))
   }
 })
 
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -68,12 +68,6 @@ jobs:
       - name: package electron - mac
         if: startsWith(matrix.os, 'mac')
         run: npm run package:mac
-      - name: npm rebuild - mac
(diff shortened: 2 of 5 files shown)


public/electron.js
```diff
@@ -2,7 +2,6 @@
 if(require('electron-squirrel-startup')) return 
 
 // Modules to control application life and create native browser window
-const imageKeyboardResponse = require('@jspsych/plugin-html-keyboard-response')
```
benjamin-heasly, line 5: I missed this in my last PR to upgrade to jspsych 7, which I didn't catch until I started testing with electron.  jspsych 7 will only run in the browser, so we can't `require()` it here on the server-side.

I'm hoping `args.trial_type` below can happily be a string, although I don't quite know how these `args` are used.

eldu (approved):
This looks good to me and it was able to build for windows in the github actions!

benjamin-heasly:
Awesome, thanks again @eldu !  I'll go ahead and merge this

PR #101, opened 2022-07-14
fix(release.yml): fix copy-paste typo in release.yml

benjamin-heasly:
Doh! I made a copy-paste error when revising release.yml!  This should fix it and fix the CI build.


eldu (approved):
🚢 

All good! Thanks for catching it !

PR #102, opened 2022-07-22
Use same startDate value between renders, for Login and JsPsychExperiment

benjamin-heasly:
Fixes a bug I introduced with jspsych 7, where Login and JsPsychExperiment got different values of
startDate, breaking Firebase document updates.

I found this while doing more testing, on a Honeycomb pilot project with the Nassar lab.


eldu (approved):
Thanks for the bug fix! 🐛

PR #105, opened 2022-08-03
Misc ideas

benjamin-heasly:
Greetings!  While working on the Nassar lab pilot task, I made a few changes that I think could also be helpful upstream here in the main template.

This PR contains 4 small changes/suggestions.

In `src/lib/utils.js`, import images as an object where the keys are familiar image names.  This would let users refer to images using familiar names like `images['myImage.png']`, as opposed to obscure webpack names like `/static/media/myImage.5dca7a2a50fb8b633fd5.png`.

I defined an env var `IMAGE_INLINE_SIZE_LIMIT=0` in a few places.  This will prevent webpack from inlining small images <10k as giant urls that contain actual encoded image data.  I'm thinking the inlining is not what we want with jsPsych, since we might be saving the image urls as part of the jsPsych trial data and stimuli.

I added a second export from `src/timelines/main.js` allowing users to provide custom jsPsych experiment options.  JsPsychExperiment will merge these with other options needed by Honeycomb.  This is different since users no longer import or init jsPsych directly.

I added a build script that allows fetching response data from Firestore, by study and participant.  It requires setting up a Firebase service account in order to authenticate. I'm thinking this could be useful to include with the template as a working example based on the same language and tooling as the rest of Honeycomb (and not have to set up a separate Python environment, for example).

What do you think?


eldu:
Looks like the builds were cancelled? Did you run into issues with building?

benjamin-heasly:
I don't think I did, huh.

Oh, looks like I have a lint error.  I'll fix...

benjamin-heasly:
Hopefully that's fixed now.  That's what I get for "cleaning up" after testing locally.

download-response-data.js
```diff
@@ -0,0 +1,115 @@
...
+ *  - Click on the "Service accounts" tab
+ *  - Near the bottom click "Generate new Private key"
+ *  - Save the key on your local machine and remember where you put it
+ *
+ * This Private key is a secret, so be careful with it!
+ * Do not share it.
+ * You may wish to restrict the file permissions on it so only you can access it.
+ * For example: "chmod 600 path/to/serviceAccountKey.json".
+ * You may also wish to delete it when you're done.
+ * Do not share it.
```
eldu: Maybe we should also add in that people shouldn't commit it into the repo too

eldu (commented):
These looks like good changes! I would have recommended having these as separate smaller PRs. No PR is too small! 

I'd like to see what others think about these changes before approving! 

Importing the images as an object with familiar names looks good! 

The `IMAGE_INLINE_SIZE_LIMIT=0` environment variable sounds good as well in terms of saving these images

As for providing custom jsPsych experiment options, this also looks like a good change however is it something that we want to allow? @mirestrepo would you know?

The build script that allows fetching data from firestore is a great way to grab data from firebase. @mirestrepo Do you know how users are getting their data from firebase at the moment? For my other projects we similarly have a script to download from firebase.

download-response-data.js
benjamin-heasly: Sounds like a good idea, I'll add that.

benjamin-heasly:
Thanks @eldu for reviewing this 4-in-1.  I think you're right, 4 would have been better.

eldu (approved):
🚀 I think these are great ideas! Would like see what others think as well

PR #106, opened 2022-08-05
Delete secret files and add to gitignore

RobertGemmaJr:
- Delete `.npmrc`
- Delete `env/.env.firebase`
- Add files to the gitignore

The docs will be reflected to make these files during the quick start and firebase setup


eldu (approved):
:shipit:

PR #108, opened 2022-08-09
refactor: add Brewfile for macOS quick installation

jashlu:


RobertGemmaJr:
I added the `chocolatey.config` file and moved both into a separate "dependency" folder but I'm not at all sure that's the best place for them.

Also, I added a few comments in the file about installing chocolatey and how to install the package, it may be good to do the same on the brewfile?

jashlu:
> I added the `chocolatey.config` file and moved both into a separate "dependency" folder but I'm not at all sure that's the best place for them.
> 
> Also, I added a few comments in the file about installing chocolatey and how to install the package, it may be good to do the same on the brewfile?

Great Idea, let me do that for the brewfile then. Also using a dependency folder would be nice for organization. We can add a line to the quick_start docs to direct users to the folder!

mirestrepo (approved):

AisOmar (approved):

PR #117, opened 2022-09-13
Add VS build tools to chocolatey.config

RobertGemmaJr:
It seems the issue people have been having in slack is because Visual Studio needs to install the c++build tools, not just the workload. I updated the choco file to do so


eldu (approved):
🚢 Thanks Rob!

PR #135, opened 2023-02-14
chore: Remove typescript dependency

### LINKED ISSUE #95: remove typescript dep if we can
I don't know why we have it, but maybe there's a reason?

RobertGemmaJr:
I wasn't able to find anything that used typescript and was able to run/build on my end. Merging into `h-3.1` as a holding branch for some small PRs that are coming soon

closes #95

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,7 @@
     "react": "^16.8.6",
     "react-bootstrap": "^1.6.1",
     "react-dom": "^16.8.6",
-    "react-scripts": "5.0.1",
-    "typescript": "^3.6.4"
+    "react-scripts": "5.0.1"
   },
   "scripts": {
     "commit": "git-cz",


digicosmos86 (approved):
Looks good to me! Why are we not using TypeScript?

RobertGemmaJr:
> Why are we not using TypeScript?

It's just never been implemented? I think there's thought to not having people who aren't familiar with JS to also have to learn TS but I don't know that for sure

mirestrepo:
Yes, no typescript here. I think JavaScript is already overwhelming to many :)On Feb 14, 2023, at 3:15 PM, Robert Gemma ***@***.***> wrote:﻿

Why are we not using TypeScript?

It's just never been implemented? I think there's thought to not having people who aren't familiar with JS to also have to learn TS but I don't know that for sure

PR #143, opened 2023-03-02
wip: Add Changes from PR #88

RobertGemmaJr:
Adds changes from #88 into the new honeycomb. There's still a LOT to do and cleanup but I want to get some eyes on it now that things are at least running.

Right now the Stroop task shows and we are able to walk through the task. I don't entirely believe that the keyboard response is working correctly (hence why this is in a holding branch) but it's no longer erroring out.

The issue with #88 is that it was written using `jspych-react` and before the project upgraded to `JsPsych 7`. I've done my best to pull out the changes and move them to the new syntax.


RobertGemmaJr:
I'm going to work on getting some of the bigger TODOs into issues but some are small enough I'll try to tackle them in the next PR.

RobertGemmaJr:
Also, it's worth noting I believe I know enough now to tackle the `startDate` and other issues so I'll work on those in separate branches that can be merged directly into main

RobertGemmaJr:
Merging into [feat-add-88](https://github.com/brown-ccv/honeycomb/tree/feat-add-88) instead of `main`

PR #144, opened 2023-03-03
fix: Fix `startDate` to update on user login

### LINKED ISSUE #126: fix: StartDate is initialized too soon for accuracy
JSPysch only saves elapsed time, for experiments that need an absolute time, this is achieved by comparing to the start date. In old honeycomb, the "login" screen was a jspysch trial that used to save the start time, now, start time happens [here](https://github.com/brown-ccv/honeycomb/blob/729770d9866cc6e82cd05e06ae630a5d67215756/src/App.jsx#L18) with is possibly way before the participant starts the experiments and therefore does not accurately represent when the experiment timeline starts

### LINKED ISSUE #128: fix: startDate needs to be Date.Now() after login
Currently, startDate refers to the app window creation, not the start of the experiment, which can be a very different time. Further, time_elapsed from jsPsych is relative to the start of the experiment, not the creation of the app window. This is problematic when synchronizing to other data collected in parallel to the task data.

### LINKED ISSUE #132: Swap `concurrently` for `wait-on`, `npm-run-all`
These have better cross platform support, should help with #123

RobertGemmaJr:
Most importantly this PR fixes the issue where a user may start the app well before they actually log in. The `startDate` is updated inside the `handleLogin` callback function before it is saved to JsPsych.

- NPM scripts use `wait-on` instead of `sleep` for better os compatability
- `startDate` is updated when the user logs in
- Splits up `validateParticipant` and `initParticipant` firebase functions
  - Participant can now be validated (`participant_id`/`study_id` combo) before creating a new experiment record in the database
- Adds some better comments in `App.jsx` and elsewhere
- Cleans up some extra bloat (e.g. makes arrow functions one liners where possible) and gives variables better names

closes #132 
closes #128 
closes #126


src/App.jsx
```diff
@@ -1,105 +1,70 @@
...
+  const [psiturk, setPsiturk] = useState(false)
+
+  // Manage the data used in the experiment
+  const [participantID, setParticipantID] = useState('')
+  const [studyID, setStudyID] = useState('')
   // Remember startDate between renders.
   // This is necessary to allow Firebase to create a timestamped document at Login time,
   // and then to find and update the *same* timestamped document after each trial in JsPsychExperiment.
-  const [startDate] = useState(new Date().toISOString())
+  // TODO: Shouldn't the document have a UID? Save the date but we should have a uid for the document?
```
eldu: Doesn't the UID get automatically generated? Or am I wrong about that?

src/App.jsx
```diff
@@ -108,35 +73,83 @@ function App () {
         window.lodash = _.noConflict()
         setPsiturk(new PsiTurk(turkUniqueId, '/complete'))
         setMethod('mturk')
-        setLoggedIn(true, 'mturk', turkUniqueId)
+        // TODO: This won't work
+        handleLogin('mturk', turkUniqueId)
```
eldu: Is there an issue made for this?

src/App.jsx
```diff
@@ -108,35 +73,83 @@ function App () {
...
         /* eslint-enable */
       } else if (config.USE_PROLIFIC) {
         const pID = getProlificId()
         if (config.USE_FIREBASE && pID) {
           setMethod('firebase')
-          setLoggedIn(true, 'prolific', pID)
+          // TODO: This wont work
+          handleLogin('prolific', pID)
```
eldu: Same question here, is there an issue for this?

src/App.jsx
```diff
@@ -108,35 +73,83 @@ function App () {
...
-        }
+        if (participantId) setParticipantID(participantId)
+        if (studyId) setStudyID(studyId)
+
+        setMethod('firebase')
       } else {
         setMethod('default')
       }
     }
+    console.log(currentMethod)
```
eldu: delete?

src/App.jsx
```diff
@@ -108,35 +73,83 @@ function App () {
...
+
+  // Default to valid
+  const defaultValidation = async () => true
+
+  // TODO: This should JUST validate, not init the participant
+  // Validate participant/study against firestore rules
+  const firebaseValidation = (participantId, studyId) => {
+    // return initParticipant(participantId, studyId, startDate)
+    return validateParticipant(participantId, studyId)
+  }
```
eldu: Can the comments here be deleted?

src/App.jsx
```diff
@@ -108,35 +73,83 @@ function App () {
...
+  const psiturkUpdateFunction = (data) => { psiturk.recordTrialData(data) }
+
+  /** EXPERIMENT FINISH FUNCTIONS */
+
+  const defaultFinishFunction = (data) => { data.localSave('csv', 'neuro-task.csv') }
+  const desktopFinishFunction = () => { ipcRenderer.send('end', 'true') }
+  const psiturkFinishFunction = () => {
+    const completePsiturk = async () => {
+      psiturk.saveData()
+      await sleep(5000)
```
eldu: Rather than this sleep statement, looks like there are callbacks.

[Source](https://psiturk.readthedocs.io/en/latest/api.html#psiturk-savedata-callbacks)

```
psiturk.saveData({
   success: function() {
      // function to run if the data is saved
   },
   error: function() {
      // function to run if there was an error
   }
});
```

eldu: Update to use the callbacks instead of sleeping

src/components/Login.jsx
```diff
@@ -1,43 +1,27 @@
...
-    .then((loggedIn) => {
-      if (loggedIn) {
-        onLogin(loggedIn, studyId, participantId);
-      } else {
-        setError(true);
-      }
+    // Logs user in if a valid participant/study id combination is given
+    validationFunction(participantId, studyId).then((isValid) => {
+      if (isValid) handleLogin(participantId, studyId);
+      else setIsError(true);
```
eldu: Suggestion, always `setIsError`? Or would we not want to for some reason here?

```suggestion
      setIsError(isValid);
      if (isValid) handleLogin(participantId, studyId);
```

eldu (changes requested):
Alright i threw in a bunch of comments! Great work rob!

src/App.jsx
RobertGemmaJr: Good catch!

src/App.jsx
RobertGemmaJr: Added #145

src/App.jsx
RobertGemmaJr: Added #145

src/App.jsx
RobertGemmaJr: Done!

src/components/Login.jsx
RobertGemmaJr: Good idea! I'll commit it

src/App.jsx
RobertGemmaJr: Somewhat unrelated but this seems to be a python package? I'm not entirely sure how we're executing the code in our project. It looks like it's passed as a script in `index.html` but why not just include it as a package?

RobertGemmaJr: I did refactor to use the callbacks

eldu (approved):
Looks good to me! Thanks rob!

src/App.jsx
mirestrepo: I think in honeycomb we use the startDate as a document identifier. But I also think that from their [best practices](https://firebase.google.com/docs/firestore/best-practices), I don't know if dates qualify as monotonically increasing and therefore if that's a good idea or not. I think we did it that way for user readability and possibly to be consistent with that is done on desktop. But be could revisit

src/components/JsPsychExperiment.jsx
```diff
@@ -25,32 +27,34 @@ function JsPsychExperiment({
...
-  // Create the instance of jsPsych that we'll reuse within the scoope of this JsPsychExperiment component.
+  // Create the instance of jsPsych that we'll reuse within the scope of this JsPsychExperiment component.
   // As of jspsych 7, we create our own jspsych instance(s) where needed instead of importing one global instance.
-  const setUpJsPsych = () => {
+  const jsPsych = useMemo(() => {
+    // Write the initial record to Firestore 
+    if(config.USE_FIREBASE) initParticipant(participantId, studyId, startDate)
+
     const jsPsych = initJsPsych(combinedOptions)
+    // Add experiment properties into jsPsych directly
```
mirestrepo, line 39: I think `startDate` is still the time of creating and not of login....

src/firebase.js
```diff
@@ -25,6 +26,24 @@ if (window.location.hostname === "localhost") {
   db.useEmulator("localhost", 8080);
 }
 
+// TODO: Can store variables for the nested pieces of a record https://firebase.google.com/docs/firestore/data-model#references
+
+// Attempts to retrieve the data in /{collectionName}/{study_id}/participants/{participant_id}
+// Will return false if participant isn't valid
+const validateParticipant = async (participantId, studyId) => {
```
mirestrepo, line 33: For online experiments there are two common situations, when participants are preregistered (small clinical trials) or when participants are not (large online prolific experiments). Maybe we should have a setting variable that controls for that?

mirestrepo (changes requested):
Thanks for diving into this! I may be missing something, but I'm not sure this addresses the `startDate` issue

mirestrepo (approved):
We chatted and my questions were clarified. The only thing we thought we could change is moving the `startDate` to the inside of the experiment, since it's no longer needed outside

src/App.jsx
RobertGemmaJr: Looks like firestore will create an ID if you don't feed it one
image

RobertGemmaJr: Added #146

src/firebase.js
RobertGemmaJr: #147

RobertGemmaJr:
I moved the initialization of `startDate` into JsPsychExperiment.jsx where we initialize the jsPsych experiment. `<JsPsychExperiment />` is only rendered after login so it's the only place we need to capture the time. I also moved the initialization of the Firestore document (the experiment itself) to the same place

PR #148, opened 2023-03-09
Style: Use `standard-js`

RobertGemmaJr:
Refactors honeycomb to use [standardjs](https://standardjs.com) instead of eslint. Also adds `lint` and `format` scripts to package.json.

Note there are a few linter warnings still because of the use of `new` in the project.


RobertGemmaJr:
There are only a few errors that still need to be handled, mostly in regards to the `new` operator

```bash
standard: Use JavaScript Standard Style (https://standardjs.com)
  ...\honeycomb\public\electron.js:2:42: Parsing error: 'return' outside of 
function (null)
  ...\honeycomb\src\App.jsx:109:24: 'PsiTurk' is not defined. (no-undef)    
  ...\honeycomb\src\lib\utils.js:88:23: 'AudioContext' is not defined. (no-undef)
  ...\honeycomb\src\trials\camera.js:14:22: 'FileReader' is not defined. (no-undef)
  ...\honeycomb\src\trials\camera.js:18:22: 'new Buffer()' was deprecated since v6.0.0. Use 'Buffer.alloc()' or 'Buffer.from()' instead. (n/no-deprecated-api)
  ...\honeycomb\src\trials\camera.js:59:32: 'MediaRecorder' is not defined. 
(no-undef)
  ...\honeycomb\src\trials\camera.js:68:28: 'Blob' is not defined. (no-undef)
```

RobertGemmaJr:
Is there a way to undo changes for a specific file all the way back to `main`/`HEAD`? Looks like I accidentally linted the min files and am not sure the best way to undo those changes 

cc @broarr

eldu:
> Is there a way to undo changes for a specific file all the way back to `main`/`HEAD`? Looks like I accidentally linted the min files and am not sure the best way to undo those changes
> 
> cc @broarr

You could make a new branch and cherry pick changes from this branch? Or at least that's how I'd approach it

PR #168, opened 2023-03-27
style: use `standard-js`

RobertGemmaJr:
Refactors honeycomb to use [standardjs](https://standardjs.com) instead of eslint. Also adds `lint` and `format` scripts to package.json.

 Right now we just ignore the linter warnings coming from the use of the `new` operator.

I'm not entirely sure what I did with git but it automatically closed the old pull request as I was trying to charry pick the changes. Everything should be all set now


mirestrepo (approved):
As far as I can see this is all just listing, and I hope it has not other effect besides making the code pretty 😅

mirestrepo:
> As far as I can see this is all just listing, and I hope it has not other effect besides making the code pretty 😅

And I like reading pretty code! I'm just hoping the `var`->`const` changes don't have a side effect we are missing, but I don't think so

RobertGemmaJr:
> I'm just hoping the `var`->`const` changes don't have a side effect we are missing, but I don't think so

That was done with the automatic formatting... I would hope/assume `standard-js` is smart enough to not change things if there was any potential of effecting the logic

PR #171, opened 2023-04-04
fix: Fix Firebase 1MiB size limit & other Firebase issues

### LINKED ISSUE #94: Upgrade firebase version
With version 9, firebase changed to a functional style of api.  It's a pretty easy translation from the OO to functional calls, but will require a few changes in `src/firebase.js`

### LINKED ISSUE #122: docs: How should Firebase be run in dev?
Use emulator? What scripts are the recommended ones from `package.json`?

### LINKED ISSUE #127: issues with download script
the download script download-response-data.js was not working. I had to replace doc.id in line 109 with sessionNumber to make it work.

### LINKED ISSUE #130: fix: Experiment data may exceed Firebase 1Mb document limit
At the moment all of the results of an experiment are saved to the same document. For some experiments, this is exceeding the 1MB limit for Firestore. We should break results into multiple documents, which may imply a deeper collection/document structure [here](https://github.com/brown-ccv/honeycomb/blob/b9e817ee6e68f4b697e7966dd94ff476032a618a/src/firebase.js#L48).

We also should look into [Firebase best practices](https://firebase.google.com/docs/firestore/best-practices) - there are references to continuous indexing that can lead to hot spotting that may be important to think about.

### LINKED ISSUE #146: Use Firestore's built in UID creation
This definitely should be done for the UID of each individual experiment. Should the `reigster_studies` and `registered_participants` have the built in UID as well?

image

RobertGemmaJr:
- Adds `firebase-service-account.json` and `participant_responses/` to gitignore as they're in my local repo
- Adds `emulator_data/` with boilerplate hierarchy
- Narrows the scope of the firestore rules to only read/write/update to the hardcoded collections
- Fixes `Login.jsx` to correctly display the error when study/participant isn't valid
- Fixes `USE_PROLIFIC` being set as `undefined` instead of false
- Adds helper functions in `firebase.js`

## Firebase Emulators

I adjusted the scripts slightly to work correctly when using Firebase.
- `firebase:dev` is now `firebase:emulators:start`
- `firebase:savedev` is now `firebase:emulators:save`
- `dev:firebase` now runs in the browser (was running in electron which ignores firebase)

I will update the docs to reflect this change. Users should run `firebase:emulators:start` in a separate terminal from `dev:firebase` while doing development.

### emulator_data

I added the boilerplate collections/documents needed by the Firebase emulators to the repo. It adds a single study (`s1`) with a single participant (`p1`).

image
image

## Firebase Size Limit

Individual trials of an experiment are added as documents in a subcollection (`{experimentID}/trials/{trialID`) instead of being appended to an array. Here `{experimentID}` is the startDate of the experiment and `trialID` is randomly generated by Firebase.

### Download Script

The trials are collected inside the download script (`download-response-data.js`) and "added back" to the json file that gets saved to the user's machine. This is to maintain compatibility with previous versions/data that users' may have.

I have changed the script slightly in the logic behind downloading data. Right now all of the experiments (sessions?) are saved when the `participantID` and `studyID` are given but not a `sessionNumber`. In the old version this would only download the latest experiment. 

**Some questions about what's best:**
- Should I reference each run through of the tasks as `experiments` or `sessions`?
- If given only `participantID` and `studyID` should I download all the experiments or only the latest?
- Should I remove `sessionNumber` entirely or keep the argument?
   - I would change `sessionNumber` to `sessionID` as the user should pass the exact ID
- If given only `studyID` should it download all sessions for all participants? (Right now this errors out) 
- Should I add extra check/confirm when download would overwrite files?

**There's a lot of commented out code in the download script because of these questions. I'll wait to merge until we decide what's best**


digicosmos86 (approved):
LGTM!

RobertGemmaJr:
Okay I refactored the download script quickly to return to the old way of downloading data.

I talked with Sienna over email and we ironed out some details about what the new script should do. I'll get that done in a different PR

- Downloading all of the experiments for a given `studyID` and `participantID`
- Downloading on a date range across multiple participants for a given `studyID`
- Better file naming conventions
- Having the startDate be the ID of a session is not neccesarily useful (we can still query by `start_time` key in the data object)

PR #176, opened 2023-04-11
ref: Refactor studyID and participantID order

### LINKED ISSUE #174: Reverse `participantID` and `studyID` order
The `studyID` is the "parent" document, it should be first wherever possible.

The input fields are also in reverse order, should I swap those as well? That's a somewhat major UI change for the project

RobertGemmaJr:
`studyID` is the parent of `participantID` so it should appear first in the function definitions but it appeared second in most of them. 

- Renames `studyId` as `studyID`
- Renames `participantId` as `participantID`
- reverses order of `studyID` and `participantID` in functions
- Removes `height` and `width` props in `<JsPsychExpierment />`, sets hard code values since they don't change

I won't change the order of the fields on the `<Login />` page yet as that's a larger UI change. Do y'all think it should be changed, though?


RobertGemmaJr:
I'm merging into `v4` instead of `main` as a holding branch for the time being

eldu (approved):
looks good to me !

PR #177, opened 2023-04-11
Consolidate css

RobertGemmaJr:
- Deletes `App.css`, everything is handled in `index.css`
- Moves `App.jsx` into the `components/` folder


eldu (approved):
I had a question about why App was moving into the component folder, but then I realized that index.js is more so the root anyway

Looks good!

PR #178, opened 2023-04-12
ci: Cache npm between workflows

### LINKED ISSUE #164: Cache `npm install` on GitHub actions

RobertGemmaJr:
- Adds `cache: npm` to workflow runs

I'm not sure if this is doing anything though? The [README](https://github.com/actions/setup-node#caching-global-packages-data) says that it looks for the `package-lock.json` file but we don't commit that to the repo? And it doesn't cache `node_modules` anyways? 

I feel like this is a good change to make for our actions but I don't think it's going to properly optimize them. 

### Cache Node Modules

**Is there a way to cache node_modules between runs?** 

Here's the current matrix. Right now every single one of these are calling `npm install` but we really only need one install per OS.
- Install on Windows
  - Build for home
    - Video on
    - Video off
  - Build for clinic
    - Video on
    - Video off
- Install on MacOS
  - Build for home (video off)
  - Build for clinic (video off)
- Install on Ubuntu
  - Build for home (video off)
  - Build for clinic (video off)

### delete-artifacts

`delete-artificats.yml` is a workflow dispatch that hasn't been run in 2 years. I just did a test run and it didn't seem to do anything? GitHub auto deletes them after 90 days.

Is this workflow safe to delete?


.github/workflows/build.yml
```diff
@@ -40,23 +43,25 @@ jobs:
         run: git config --global url."https://github.com/".insteadOf "git://github.com/"
 
       # Set up js package
+      # TODO: Update these versions? @v3 (all workflows)
```
eldu, line 46: It would be good to update all of these versions! Do you run into any troubles?

eldu (approved):
I added in one comment about updating some of the versions. This pr looks pretty straightforward though ! Mostly comments and caching npm

.github/workflows/build.yml
RobertGemmaJr: We have to use an old node version (that we're looking to update), I imagine will update everything at the same time

PR #181, opened 2023-04-19
style: Add `eslint` with `prettier`

### LINKED ISSUE #167: Pre-commit hook for `standard-js`

RobertGemmaJr:
I changed my mind again. I really like `prettier`'s integration with vs code and eslint seems to be the easiest way to get it running. This also lets us add back the eslintrc file that used to be in the repo. 

I added a `.prettierrc.js` file with some of my personal preference rules but am open to deleting it (would use Prettier's default rules) if we think it's best to hide them from honeycomb users


eldu (approved):
I was thinking about this yesterday, and as for `.prettierrc.js` I can see how using the defaults would be a good idea especially for beginners. Though looking at the changes you've made it doesn't seem like that much has changed. 

Maybe adding in a comment to explain what `.prettierrc.js` is and what the linting does?

RobertGemmaJr:
> Maybe adding in a comment to explain what `.prettierrc.js` is and what the linting does?

I added a comment at the top of the file and some inline comments for the specific rules. Thanks Ellen!!

PR #182, opened 2023-04-19
ref: Use function syntax over arrow functons

RobertGemmaJr:
- Use `function` syntax instead of `() =>` for any functions longer than a single line.
- `export function` inline, not at the end of files

I decided to keep the functions in `App.jsx` as arrow functions because they're only one line and will be refactored into different files soon anyways.


eldu:
What was the thought process for using function syntax over arrow function syntax?

mirestrepo:
> What was the thought process for using function syntax over arrow function syntax?

I think one of the considerations was to make it more readable for the community that works with these tasks - since most of them are new to javascript

PR #188, opened 2023-05-18
chore: Fix warning for url to ccv website

RobertGemmaJr:
I also re-ran the `firebase init hosting:github` which updated some versioning of the github actions


package.json
```diff
@@ -4,7 +4,7 @@
   "author": {
     "name": "Brown CCV",
     "email": "ccv-bot@brown.edu",
-    "url": "ccv.brown.edu"
+    "url": "http://ccv.brown.edu"
```
eldu:
```suggestion
    "url": "https://ccv.brown.edu"
```

eldu (approved):
I suggested changing the url to have `https`

package.json
RobertGemmaJr: Good catch! Thanks!

PR #189, opened 2023-06-12
ref: Use `winget` not `choco`

### LINKED ISSUE #134: Use `winget` instead of `chocolatey` for setting up Windows dev environemnt
`winget` is a package manager directly from Windows. Not everything is available yet but they're adding stuff pretty quickly

RobertGemmaJr:
- Adds PowerShell script file for winget installation
- Removes chcolatey file
- Adds oracle JDK to install list 

Note that this PR relies on the 3.1 documenation PRs ([here](https://github.com/brown-ccv/honeycomb-docs/pull/54), [here](https://github.com/brown-ccv/honeycomb-docs/pull/68), and [here](https://github.com/brown-ccv/honeycomb-docs/pull/70)). Everything should be merged at the same time


Brewfile
```diff
@@ -1,8 +1,11 @@
...
-cask "visual-studio-code"
+# Brewfile for installing Honeycomb prerequisites
+# Visit https://brown-ccv.github.io/honeycomb-docs/docs/quick_start for info
+
+brew "git"
+brew install --cask "github"
+brew install "node@18"
+brew install --cask "oracle-jdk"
+brew install --cask "visual-studio-code"
+
```
mirestrepo, line 9: No need for firebase-cli? Or did you move it as a node dee dependency?

Why do we need JDK?

RobertGemmaJr: Yeah, it's an `npm i -g` install now. That's the only way to get it on Windows so I figured we'd keep it the same for both.

The JDK is for the Firestore Emulators.  It runs some java based backend when using the emulators locally

PR #190, opened 2023-06-13
ref: Use function syntax, not arrows

### LINKED ISSUE #160: Use function syntax instead of arrow functons
If there's one liner functions it's okay to leave them as arrows - just add a comment about how arrow syntax works. All the utility functions should definitely be written as normal functions with comments

RobertGemmaJr:
Completes what was started in #182, but ONLY refactors the function stuff, will mess with the `export` declerations in another PR.

- Use function syntax instead of arrows
- Adds documentation to functions
- Generally cleans up comments


mirestrepo (approved):
Looks good to me!

PR #202, opened 2023-06-21
ref: Seperate JsPsych folder

RobertGemmaJr:
Right now I'm copying folders over where I can because some of the old code still uses them. Once I move over all of the trials I will delete the files outside of `src/JsPsych`. Large PR 😬 but mostly just moving things around, I think the directory structure of the "Files Changed" tab gives the best impression of what's changed. Lots more to do but want to get some eyes before the PR gets completely out of hand.

- Adds `JsPsych/` folder inside src
- Copies assets used by the experiment from `assets/images` to `JsPsych/assets/images`
- **Adds example JSON files for configuration to `JsPsych/config**
    - `clinic.json` and `home.json`
- Adds `JsPsych/language` folder with ONLY `en_US.json`
  - Adds an `index.js` file for loading that file, perhaps this becomes i18n at some point?
-  Adds `preamble` as an example trial in `JsPsych/trials/exampes` as `Preamble.js`
- Moves `holdUpMarker` and `startCode` to `JsPsych/trials/examples/photodiode` as `HoldUpMarker` and `StartCode`
- Adds `JsPsych/constants.js` file for constant values
    - These were in `/config/main` before. I've only copied the values I'm using thus far
- Adds `JsPsych/timeline.js` for building the timeline and setting the JsPsych options
   - This is the only file that the end user will be expected to have? These exports are what will be passed as  a prop to the package
-  Adds `JsPsych/utils.js` for utility functions
     - These were in `/lib/utils.js` before. I've copied all of the functions as they're only used in the building of jspsych stuff


RobertGemmaJr:
I changed the way I export the trials a little bit in #203

src/JsPsych/timeline.js
```diff
@@ -0,0 +1,43 @@
...
+    // taskBlock(practiceBlock),
+    // countdown({ message: lang.countdown.message2 }),
+    // taskBlock(exptBlock1),
+    // demographics,
+    // iusSurvey,
+    // debrief,
+    // showMessage(config, {
+    //   duration: 5000,
+    //   message: lang.task.end,
+    // }),
```
eldu, line 37: is this part of a todo? or can there be an explanation of this to the user?

src/JsPsych/trials/examples/Preamble.js
```diff
@@ -0,0 +1,61 @@
+import htmlKeyboardResponse from '@jspsych/plugin-html-keyboard-response';
+import htmlButtonResponse from '@jspsych/plugin-html-button-response';
+import { showMessage } from '@brown-ccv/behavioral-task-trials';
+
+// import holdUpMarker from '../trials/holdUpMarker';
+// import startCode from '../trials/startCode';
+// import { lang, config } from '../config/main';
```
eldu, line 7: More commented code ;)

eldu (approved):
passing through! some commented code

no need to change the docstrings you have here, just wanted to throw out there that I like google's doc strings. The following example is in python

```python
  def connect_to_next_port(self, minimum: int) -> int:
    """Connects to the next available port.

    Args:
      minimum: A port value greater or equal to 1024.

    Returns:
      The new minimum port.

    Raises:
      ConnectionError: If no available port is found.
    """
    if minimum < 1024:
      # Note that this raising of ValueError is not mentioned in the doc
      # string's "Raises:" section because it is not appropriate to
      # guarantee this specific behavioral reaction to API misuse.
      raise ValueError(f'Min. port must be at least 1024, not {minimum}.')
    port = self._find_next_open_port(minimum)
    if port is None:
      raise ConnectionError(
          f'Could not connect to service on port {minimum} or higher.')
    assert port >= minimum, (
        f'Unexpected port {port} when minimum was {minimum}.')
    return port
```

https://google.github.io/styleguide/pyguide.html

src/JsPsych/timeline.js
RobertGemmaJr: Oh yeah it's worth noting to anyone looking at the PRs that these are all going into a `v4` holding branch, not main. So all the commented out code is things to do that I'm still working on

RobertGemmaJr:
> no need to change the docstrings you have here, just wanted to throw out there that I like google's doc strings. 

Here's what they have for [js](https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments)! I'll definitely go through and clean things up later on

PR #203, opened 2023-06-22
ref: Move `ageCheck` and `sliderCheck` into `JsPsych` folder

RobertGemmaJr:
This is merging into #202 

- Moves `ageCheck.js` to `JsPsych/trials/examples/surveys` folder as `AgeCheck.js`
- Moves `sliderCheck.js` to `JsPsych/trials/examples` folder as `Slider.js`
- Updates how the trials are exported (see below)

I've updated how I'm exporting the trials and I think this makes more sense. The function is explicitly named as `create_` to denote that it's a function for building the JsPsych trial object. I also export a constant that just calls the function. This way the trials can be added neatly when building the timeline but you can still use the function if you need to pass parameters.

```js
  // Get slider text from the language file and create the trials
  const sliderMessages = language.quiz.direction.slider;
  const sliderLeft = createSliderTrial(sliderMessages.left);
  const sliderRight = createSliderTrial(sliderMessages.right);
  const timeline = [
    Preamble, // Add the trial (imported directly)
    AgeCheck,  // Add thetrial (imported directly)
    sliderLeft,  // Add the trial (built above)
    sliderRight,  // Add the trial (built above)
    ...
]
```


galenwinsor (approved):
I like this pattern! It does sort of obfuscate the logic a little, which could potentially lead to some issues, I suppose, but I certainly like the way it looks in declaring the timeline. Isa mentioned to me that we might be transitioning Honeycomb to a library rather than a template at some point. In that case, I wonder if this pattern would still make sense. But that's probably far off!

RobertGemmaJr:
> Isa mentioned to me that we might be transitioning Honeycomb to a library rather than a template at some point. 

All of the code I'm moving is the code that will NOT be in the library. It's all code for creating a JsPsych experiment which will get passed to the (eventual) honeycomb library's component.

It's definitely a bit weird while we're in this middle step but I'm hoping to get to the point where I can mentally ignore the `JsPsych` folder while I work on the library but still be able to run the app like normal

PR #205, opened 2023-06-22
ref: Refactor countdown trial

RobertGemmaJr:
- Adds `Countdown.js` example trial, copies code from [trials repo](https://github.com/brown-ccv/behavioral-task-trials/blob/main/trials/countdown.js)

### DIFF
--- a/src/JsPsych/trials/examples/Countdown.js
+++ b/src/JsPsych/trials/examples/Countdown.js
@@ -1,0 +1,36 @@
+import { range } from 'lodash';
+import htmlKeyboardResponse from '@jspsych/plugin-html-keyboard-response';
+
+/**
+ * Builds a countdown transition with the given message and number of seconds.
+ * Counts down from "time" to 0, each step is displayed for "duration" seconds
+ *
+ * @param {number} duration - trial duration in milliseconds. (default: 1000)
+ * @param {string} stimulus - Onscreen stimulus in HTML to be shown in the trial. If the stimulus is not provided, message should be provided as a string. (default: "")
+ * @param {string} message - (optional) message for the countdown. (default: "")
+ * @param {number} time - start number for the countdown. (default: 3)
+ */
+// TODO: Always use <h3>? Pass message and remove stimulus? Default to undefined instead of empty strings?
+// TODO: Can we do this in 1 trial instead of a nested timeline?
+// TODO: This isn't a keyboard response? Just instructions? Definitely don't need user input
+export function createCountdownTrial({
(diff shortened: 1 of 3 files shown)


src/JsPsych/trials/examples/Countdown.js
```diff
@@ -0,0 +1,36 @@
...
+// TODO: Always use <h3>? Pass message and remove stimulus? Default to undefined instead of empty strings?
+// TODO: Can we do this in 1 trial instead of a nested timeline?
+// TODO: This isn't a keyboard response? Just instructions? Definitely don't need user input
+export function createCountdownTrial({
+  duration = 1000, // Default to 1 second per trial
+  stimulus = '', // Default empty stimulus
+  message = '', // Default empty message
+  time = 3, // Default 3...2...1... countdown
+} = {}) {
+  const stimulusOrMessage = message !== '' ? `<h3>${message}</h3>` : stimulus;
```
galenwinsor, line 22: I'm a little confused by this fallback, since both are empty strings by default. Do you want to require the caller to provide one? I guess this is JS 🙄  so it might be hard.

src/JsPsych/trials/examples/Countdown.js
```diff
@@ -0,0 +1,36 @@
...
+  return {
+    type: htmlKeyboardResponse,
+    stimulus: stimulusOrMessage,
+    trial_duration: duration,
+    response_ends_trial: false,
+    timeline: timeline,
+  };
+}
+
+export const Countdown = createCountdownTrial();
```
galenwinsor, line 36: I might be missing something, but why are there two exports?

galenwinsor (approved):
Looks good to me, just those two points of confusion but I'm not super familiar with Honeycomb anymore so you might just want to ignore me 😅

src/JsPsych/trials/examples/Countdown.js
galenwinsor: Nevermind, I get it! Elegant!

src/JsPsych/trials/examples/Countdown.js
RobertGemmaJr: I think I'm going to remove the `simulus` parameter and just have the caller pass `message`. This way the function always adds the `<h3` markup and the caller never has to worry about html

anna-murphy: This change feels like a good direction to me.

anna-murphy (approved):
This feels nice and clean to me. I don't have any major comments here -- it looks good!

PR #216, opened 2023-06-27
Ref welcome

RobertGemmaJr:
Still in the midst of moving things around so theirs lots of commented out code and such - please ignore!

- Deletes `quizTrial` trial (moved in #214)
- Splits the two trials in `welcome` into `WelcomeMessage` and `LargeWindow`

Note that JsPsych now has a [fullscreen plugin](https://www.jspsych.org/7.0/plugins/fullscreen/) so I'll eventually implement that and delete `LargeWindow`


digicosmos86 (approved):
Looks great to me!

PR #217, opened 2023-06-27
ref: Move `camera.js` into `trials/examples/camera`

RobertGemmaJr:
Still in the midst of moving things around so theirs lots of commented out code and such - please ignore!

- Splits `camera.js` into `CameraStart` and `CameraEnd` trials

Note that JsPsych now has an [initialize-camera plugin](https://www.jspsych.org/7.3/plugins/initialize-camera/)! Will implement that and delete these trials later. Plus it won't be constrained to home use!

### DIFF
--- a/src/JsPsych/timeline.js
+++ b/src/JsPsych/timeline.js
@@ -1,20 +1,9 @@
 // TODO 204: @ import for language
 import { language } from './language';
-// TODO 204: @ import for trials
-import {
-  Preamble,
-  createCountdownTrial,
-  createSliderTrial,
-  AgeCheck,
-  EndExperiment,
-  AdjustVolume,
-  LargeWindow,
-  WelcomeMessage,
-} from './trials/examples';
-import { createHoneycombBlock } from './trials/honeycombBlock';
 
-import { DebriefSurvey, IusSurvey } from './trials/examples/survey';
-import { DemographicsSurvey } from './trials/examples/survey/Demographics';
+// TODO 204: @ import for trials
+import { Preamble, createCountdownTrial, EndExperiment } from './trials/examples';
+import { createHoneycombBlock } from './trials/honeycombBlock';
 
 /**
  * Create your custom JsPsych options here. These settings will applied experiment wide.
@@ -39,9 +28,9 @@ export const JSPSYCH_OPTIONS = {
 // eslint-disable-next-line
 export function buildTimeline(jsPsych) {
   // Get slider text from the language file and create the trials
-  const sliderLanguage = language.quiz.direction.slider;
-  const sliderLeft = createSliderTrial(sliderLanguage.left);
-  const sliderRight = createSliderTrial(sliderLanguage.right);
+  // const sliderLanguage = language.quiz.direction.slider;
+  // const sliderLeft = createSliderTrial(sliderLanguage.left);
+  // const sliderRight = createSliderTrial(sliderLanguage.right);
 
   // Get countdown txt from the language fil and create the trials+
   const countdownLanguage = language.countdown;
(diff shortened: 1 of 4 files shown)


src/JsPsych/timeline.js
```diff
@@ -39,9 +28,9 @@ export const JSPSYCH_OPTIONS = {
 // eslint-disable-next-line
 export function buildTimeline(jsPsych) {
   // Get slider text from the language file and create the trials
-  const sliderLanguage = language.quiz.direction.slider;
-  const sliderLeft = createSliderTrial(sliderLanguage.left);
-  const sliderRight = createSliderTrial(sliderLanguage.right);
+  // const sliderLanguage = language.quiz.direction.slider;
+  // const sliderLeft = createSliderTrial(sliderLanguage.left);
+  // const sliderRight = createSliderTrial(sliderLanguage.right);
```
anna-murphy, line 33: Worth it to delete these comments?

src/JsPsych/trials/examples/camera/CameraEnd.js
```diff
@@ -0,0 +1,40 @@
...
+
+// TODO: Move markup to JsPsych?
+import { photodiodeGhostBox } from '../../../../lib/markup/photodiode';
+import { baseStimulus } from '../../../../lib/markup/stimuli';
+
+import { language } from '../../../language';
+
+// TODO: This is a task, how do I pass which config file to use?
+// Hard code for now
+import config from '../../../config/home.json';
```
anna-murphy, line 11: What we do in some PPHC apps (esp Profound) is that we import all the JSON files we *might* need, and pass in a string argument to to the exported function (here `createCameraEndTrial`), and then write a new function that would take that string argument and the right JSON file. For example, it ends up looking something like this:

```js
import jsonOne from './jsonOne.json';
import jsonTwo from './jsonTwo.json';
import jsonThree from './jsonThree.json';

function selectJson(jsonKey /*string*/) {
   if (jsonKey === 'one') return jsonOne;
   // etc
}

export function createCameraEndTrial(duration, jsonKey /* still a string */) {
  const theJsonFileYouWanted = selectJson(jsonKey);
  // Whatever else you needed
}
```

This could definitely be a future kind of upgrade though.

src/JsPsych/trials/examples/camera/CameraEnd.js
```diff
@@ -0,0 +1,40 @@
...
+export function createCameraEndTrial(duration) {
+  const stimulus =
+    baseStimulus(`<h1>${language.task.recording_end}</h1>`, true) + photodiodeGhostBox();
+
+  return {
+    type: htmlKeyboardResponse,
+    stimulus,
+    trial_duration: duration,
+    on_load: () => {
+      //   if (config.USE_CAMERA) {
```
anna-murphy, line 26: Re, delete comments?

src/JsPsych/trials/examples/camera/CameraStart.js
```diff
@@ -0,0 +1,124 @@
...
+// TODO 192: Add warning to trial if not running in electron?
+let ipcRenderer = false;
+// TODO: Camera can now be accessed by the browser
+// if (config.USE_ELECTRON) {
+// if (config.equipment.camera === true) {
+//   ipcRenderer = window.require('electron').ipcRenderer;
+// }
+
+// ! This will break, need only a single JsPsych instance for the whole app
+const JSPSYCH = initJsPsych();
```
anna-murphy, line 26: Is there any reason this couldn't be passed into `createCameraStartTrial`? Or, if it can't, would it be worth exporting a standalone JSPSYCH object from a different file so it can be important globally?

anna-murphy (approved):
This looks okay to me, I just added a few notes on your code, but because this is part of a larger process, feel free to ignore them if they're going to be addressed later.

src/JsPsych/timeline.js
RobertGemmaJr: Deleted in #218!

src/JsPsych/trials/examples/camera/CameraStart.js
RobertGemmaJr: With JsPsych 7 you now have  a single, "central", JsPsych instance like you're talking about. Some of these trials were written  before that which is why they're spinning up their own instances 😞 

Moreover, we initialize our JsPsych instance inside the `<JsPsychExperinment />` which is _after_ the end user will create their trials. It's definitely an issue we'll have to come up with a more elegant solution for

anna-murphy: Ugh, don't like that... Feel free to resolve this then. If you want to brainstorm a better way to set that up, I'd be happy to help, but that's probably outside the scope of this issue (and I'm sure you'll figure out something good anyway).

PR #242, opened 2023-07-06
feat: Update react to v18

RobertGemmaJr:
- Updates `react` to 18.2.0
- Updates `react-dom` to 18.2.0
- Updates `react-bootstrap` to  2.8.0
- Adds `package-lock.json` back into the repo

PR #245, opened 2023-07-06
feat: Update actions to use Node 18

### LINKED ISSUE #219: Use Node 18
Need to update the dependencies first. Looks like the node version is mostly specified in the workflows?

Is it worth specifying the node environment within the project as a whole? At least for the repo the end user will have? That would help us control what version they're using

### LINKED ISSUE #240: GitHub action versions needed to be updated
- actions/checkout@v1 now has @v3
- actions/setup-node@v1 now has @v3

RobertGemmaJr:
- Updates actions to use `v3` for checkout and setup
- Updates actions to use `ci` now that `package-lock.json` is in the repo

This PR relies on #242


digicosmos86 (approved):
Looks good to me! Looks like the react versions indeed were due for an update!

PR #251, opened 2023-07-13
Reverse participantID and studyID in all functions with the former coming before the latter

### LINKED ISSUE #249: Switch between `pid` and`'prolific'` in `handleLogin`
https://github.com/brown-ccv/honeycomb/blob/fb318ae4380094c40fa7558da8ceef9557877669/src/App.jsx#L78

The above line sets `prolific` as participant_id and `prolific_id` as `study`. It makes more sense to switch it around

digicosmos86:
There are a few functions with `participantID` coming before `studyID`, which is inconsistent with the rest of the code base and illogical. This PR reverses the their orders in the signature of such functions.

Also updated `prettier` and added `@babel/plugin-proposal-private-property-in-object` to squash the following warning message:

```
One of your dependencies, babel-preset-react-app, is importing the
"@babel/plugin-proposal-private-property-in-object" package without
declaring it in its dependencies. This is currently working because
"@babel/plugin-proposal-private-property-in-object" is already in your
node_modules folder for unrelated reasons, but it may break at any time.

babel-preset-react-app is part of the create-react-app project, which
is not maintianed anymore. It is thus unlikely that this bug will
ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to
your devDependencies to work around this error. This will make this message
go away.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "homepage": ".",
   "repository": "https://github.com/brown-ccv/honeycomb",
   "dependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
     "@brown-ccv/behavioral-task-trials": "^2.0.0",
     "@fortawesome/fontawesome-free": "^5.9.0",
     "@jspsych/plugin-html-button-response": "^1.1.1",
@@ -50,7 +51,7 @@
     "electron": "19.0.8",
     "eslint": "^8.38.0",
     "eslint-config-prettier": "^8.8.0",
-    "eslint-plugin-prettier": "^4.2.1",
+    "eslint-plugin-prettier": "^5.0.0",
     "firebase-admin": "^11.0.1",
     "firebase-tools": "^11.0.1",
     "husky": "^8.0.3",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
   "homepage": ".",
   "repository": "https://github.com/brown-ccv/honeycomb",
   "dependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
```
RobertGemmaJr: Can you make this a dev dependency? Looks like that is what the warning is calling for

RobertGemmaJr (changes requested):
Awesome! Thank you so much for this! Only requesting changes to get that dependency moved to a `devDependency`. Otherwise everything looks great!

package.json
digicosmos86: Yep. Will move it

RobertGemmaJr (approved):
Awesome! Thanks again! Once the actions finish feel free to merge (:

anna-murphy (approved):
Great work!

PR #254, opened 2023-07-19
ref: Refactor `useConfig`

### LINKED ISSUE #237: Move `useOldConfig` into `utils`

RobertGemmaJr:
- Pulls `useOldConfig` into a utility file and use throughput the app
- Removes env file and scripts for `turk-prolific`
- Removes `ReactDOMRender` in favor of `createRoot` (error from React 18)
- Render index in `<StrictMode>`


src/JsPsych/trials/examples/partials/photodiode.js
```diff
@@ -33,7 +18,7 @@ try {
   }
 } catch (e) {
   console.warn('window.require is not available');
-  console.error(e);
+  // console.error(e);
```
eldu: uncomment or delete?

src/components/Login.jsx
```diff
@@ -45,7 +45,7 @@ function Login({
       </Form.Group>
       <Button
         className='width-50'
-        block
+        block='true'
```
eldu: the string true?

eldu (approved):
I added in two comments, this looks great though !

src/components/Login.jsx
RobertGemmaJr: Was getting a weird error in the console about it not supposed to be a boolean... just removed it and it seems to be working fine though? I'll leave it deleted

hetd54 (approved):

PR #255, opened 2023-07-19
style: Use double quotes

RobertGemmaJr:
Updates the prettier rules and runs an `npm run format`


anna-murphy (commented):
I just did a quick `grep -rn "'" src/` to make sure you didn't miss anything, and I found a few single quotes (outside of comments and not being used as apostrophes). You didn't make any changes there, so I can't suggest in GitHub, but its in `src/components/App.jsx` and a bunch of instances under `src/JsPsych/trials`. It might be worth double checking you're linting file is hitting all the places you want?

RobertGemmaJr:
> and a bunch of instances under `src/JsPsych/trials`. 

These are all inside template literal strings so they're not being formatted correctly. I think that's something I'll work on as time goes on? Most of the trials won't be used in the next version of Honeycomb so I think it's safe to leave them. I did go through and catch some more to change though - thank you!

PR #256, opened 2023-07-20
feat: Move to new environment variables

### LINKED ISSUE #246: Rename `package:[env}` scripts to `make:`
The scripts are using `electron-forge`s `make` command, not `package`. Should reflect as such

RobertGemmaJr:
Refactors all of the old environment variables into `LOCATION` `DEPLOYMENT`. `LOCATION` is "home" or "clinic" which delineates running on the browser or electron respectfully. `DEPLOYMENT` sets which data updating/saving functions Honeycomb is going to use. "csv" is the old default, which saves the data as a csv at the end of the experiment. "local" is the current default when running in Electron - it saves a JSON in a nested folder structure on the desktop. "firebase" wires up the firebase functions. There are a couple other options and more that we could add! For now we default to `firebase`.

That said the `OLD_CONFIG` is still very much in use because it's so intertwined with the application. I've untangled the JSX components from it in this PR but there is plenty of work still to go...

### Env files

Note that the `clinic` environment files set `BROWSER=none` so the app opens in the Electron window and not the browser

- `.env.dev.clinic` is used by `npm run dev:clinic`
- `.env.dev.home` is used by `npm run dev:home`
- `.env.production.clinic` is used by `npm run build:clinic`
- `.env.production.home` is used by `npm run build:home`

### Scripts

- The `dev:*` scripts use `dotenv` to set the environment for the given `start:*` script
- The `start` and `build` scripts now only call `react-scripts` 
- Removes `*:video` scripts
- Removes the `*:firebase` sciprts
- Renames the `package:*` scripts as `make:*`
- Puts the scripts in alphabetical order

### GH Actions

- Updates the actions to call the correct scripts
- Updates some of the action versions
- Renames `setting` matrix as `location` in all workflows
- Renames a bunch of the workflow steps
- Actions no longer use "xom9ikk/dotenv@v1.0.2" to load the environment variables - they call the correct script
- Renames `build.yml` as `build_and_test.yml`
- Build workflow no longer adds `video` on/off to the matrix
- Build and release workflow now run `npm run lint`

### Misc.

- `App.jsx`, `Login.jsx` and `JsPsychExperiment.jsx` now use the new environment variables for their logic
- Adds a new `constants` and `utils` files a the base of `src/`
- Moves Firebase config into its own JSON file
- Moves Firebase files into their own `src/firebase/` folder
- Deletes the `home.json` and `clinic.json` files
- Deletes `useOlfConfig` function in favor of `OLD_CONFIG` constant
- `OLD_CONFIG` is now (partially) set by the environment variables
- Moves `getQueryVariable` out of JsPsych utils folder and into base `utils` file
- Deletes `getProlificID` utility function
-


src/firebase/firebase.config.json
broarr, line 1: Should this be committed?

broarr (approved):
Looks good to me! Nice way to refactor this stuff

src/utils.js
```diff
@@ -13,3 +13,18 @@ export function useOldConfig(newConfig) {
...
+/**
+ * Get a query parameter out of the window"s URL
+ * @param {*} variable The key to parse
+ */
+// TODO 199: Can this just use URLSearchParams?
+// TODO: MOve this out of JsPsych
+export function getQueryVariable(variable) {
+  const query = window.location.search.substring(1);
+  const vars = query.split("&");
+  for (let i = 0; i < vars.length; i++) {
```
hetd54, line 26: Look at forEach Array method -- I think it makes sense to use here if you think it will make the code more readable. Obv just a style choice though, feel free to ignore!

src/utils.js
```diff
@@ -13,3 +13,18 @@ export function useOldConfig(newConfig) {
     USE_CAMERA: equipment.camera ? true : false,
   };
 }
+
+/**
+ * Get a query parameter out of the window"s URL
+ * @param {*} variable The key to parse
+ */
+// TODO 199: Can this just use URLSearchParams?
+// TODO: MOve this out of JsPsych
```
hetd54, line 22: Do you need this todo?

.github/workflows/release.yml
```diff
@@ -2,103 +2,95 @@ name: release
...
+        run: npm run make:mac
 
       # Upload installers to github release
-      - name: Upload app to release - windows
+      - name: 🚀 Upload Installer to Release (Windows)
         if: startsWith(matrix.os, 'windows')
         uses: svenstaro/upload-release-action@v2
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: out/make/squirrel.windows/x64/${{ steps.package_info.outputs.package_name }}-${{ steps.package_info.outputs.package_version }} Setup.exe
```
hetd54, line 72: "Setup.exe" here, but "setup.exe" elsewhere

.github/workflows/package.yml
```diff
@@ -4,17 +4,18 @@ name: package
 on:
   workflow_dispatch:
     inputs:
-      setting:
+      location:
+        # TODO: Home doesn't make sense here?
```
hetd54, line 8: Maybe remove this TODO since the default changed from "home" to "clinic" --

btw, you have in the release yml:
> location: ["home", "clinic"]

and no default -- is that intended to be different here?

hetd54 (approved):
Just passing through, so someone with more react/github-actions experience should take a look as well! I have a couple of comments but nothing major. Looks good to me!

src/firebase/firebase.config.json
RobertGemmaJr: Yup! This is the client side stuff that's okay to commit

src/utils.js
RobertGemmaJr: Yeah this is an old function... I'm thinking [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) are in order but I'm not sure if they work with Electron yet

.github/workflows/package.yml
RobertGemmaJr: `package.yml` is a manual workflow in which you pick a setting, which is why there's a default. `release.yml` runs both home and clinic automatically! I'll leave the TODO because it should probably be fully deleted on the `package.yml` workflow. I just don't want to touch it while I'm still moving files around

src/utils.js
RobertGemmaJr: Good catch! Thanks 😄

PR #257, opened 2023-08-04
feat: Dynamically import deployment functions

### LINKED ISSUE #157: Move validation, update, & finish functions into their own files
Use `MODE` from config.json to determine which functions to actually load and pass into JsPsych.

- electron.js
- firebase.js
- psiturk.js

RobertGemmaJr:
The CRUD functions are pulled into separate folders and dynamically imported via the new `import([path])` syntax

- Moves CRUD functions to `deployments/[deployment]`
- Renames the deployment function to exactly match the jsPsych function they're replacing
- The `deployment` state variable is now the module being dynamically imported
- Adds a `getDeployment` utility function
- Moves `fullscreen` function into nested JsPsych folder (I think this was a merge issue?)
- Adds a `isElectron` utility function
- Updates to ECMA 2020
- Fixes import issue with the min files


package.json
```diff
@@ -49,7 +49,7 @@
     "@electron-forge/maker-zip": "^6.2.1",
     "@electron-forge/plugin-auto-unpack-natives": "^6.2.1",
     "babel-loader": "^9.1.2",
-    "babel-plugin-macros": "^2.6.1",
+    "babel-plugin-macros": "^2.4.2",
```
broarr, line 52: Why the downgrade?

src/components/App.jsx
```diff
@@ -64,7 +59,7 @@ function App() {
         setParticipantID(credentials.participantID || "");
         setStudyID(credentials.studyID || "");
 
-        setIpcRenderer(renderer);
+        // setIpcRenderer(renderer);
```
broarr: You don't need this?

broarr (approved):
This looks good to me!

package.json
RobertGemmaJr: It caused a weird conflict with another peer dependency - I think it was `fs` and `fs-extra`?

src/components/App.jsx
RobertGemmaJr: Nope! Good catch!

PR #258, opened 2023-08-04
feat: Add `nvm`

### LINKED ISSUE #252: Use and enforce node version with `nvm`
- [ ] Install `nvm` during install script, install LTS
- [ ] Add `nvmrc` file to set/install/use node version within the honeycomb project

Is there an easy way to make the actions match as well? They're using strings but may be able to load from a local nvmrc file?

RobertGemmaJr:
- Adds `.nvmrc` file
- Installs `nvm` instead of node on Windows and mac
- Fixes the `Brewfile`
- Removes comments about the downloading data script (now in Honeycomb docs [here](https://github.com/brown-ccv/honeycomb-docs/pull/76)

Mac friends - all you need after installing `nvm` is to do a `nvm install [version]` and `nvm use`, right? Will be updating the documentation repo as well


Brewfile
```diff
@@ -1,11 +1,9 @@
 # Brewfile for installing Honeycomb prerequisites
 # Visit https://brown-ccv.github.io/honeycomb-docs/docs/quick_start for info
-
 brew "git"
-brew install --cask "github"
-brew install "node@18"
-brew install --cask "oracle-jdk"
-brew install --cask "visual-studio-code"
+brew  "nvm"
```
hollandjg: Looks like you have an extra space here: 
```suggestion
brew "nvm"
```

Brewfile
```diff
@@ -1,11 +1,9 @@
...
 # Visit https://brown-ccv.github.io/honeycomb-docs/docs/quick_start for info
-
 brew "git"
-brew install --cask "github"
-brew install "node@18"
-brew install --cask "oracle-jdk"
-brew install --cask "visual-studio-code"
+brew  "nvm"
+cask "github"
+cask "oracle-jdk"
```
hollandjg: Can we use the Oracle JDK any more? Wasn't there a thing to do with the license, that we shouldn't be using any Oracle Java distributions? (Here was the announcement: https://today.brown.edu/announcements/158255)

hollandjg:
> Mac friends - all you need after installing `nvm` is to do a `nvm install [version]` and `nvm use`, right?

Installing is a bit more manual – this is from the homebrew info:
```
==> Caveats
Please note that upstream has asked us to make explicit managing
nvm via Homebrew is unsupported by them and you should check any
problems against the standard nvm install method prior to reporting.
```

So perhaps we should point people to run the install script [linked on the nvm README](https://github.com/nvm-sh/nvm) instead.

.nvmrc
```diff
@@ -0,0 +1 @@
+v18.17.0
```
hollandjg, line 1: This actually just worked when using [`rtx`](https://github.com/jdxcode/rtx) – might it also work straight away with `asdf` too? Maybe it's because I already had the node plugin installed, though.

Brewfile
hollandjg:
```suggestion
brew "openjdk"
```

hollandjg (commented):
Thanks for asking for the review! I've got some questions on the actual dependencies and the NVM installation. Not sure what the right answer is though!

Brewfile
RobertGemmaJr: Will change!

RobertGemmaJr:
> > Mac friends - all you need after installing `nvm` is to do a `nvm install [version]` and `nvm use`, right?
> 
> Installing is a bit more manual – this is from the homebrew info:
> 
> ```
> ==> Caveats
> Please note that upstream has asked us to make explicit managing
> nvm via Homebrew is unsupported by them and you should check any
> problems against the standard nvm install method prior to reporting.
> ```
> 
> So perhaps we should point people to run the install script [linked on the nvm README](https://github.com/nvm-sh/nvm) instead.

Ugh... yeah this is something we're really trying to avoid. I suppose if nvm is the only tool we need to install manually it's not the end of the world though.

PR #260, opened 2023-08-09
ref: Rename "home" as "online"

RobertGemmaJr:
Refactors the location known as "home" to be "online". I get what "home" and "clinic" use is trying to get at but I think the location we're referring to here is the location of the deployment. Honeycomb is either being deployed on a "clinic" machine, or "online" where anyone can access the task from home. 

I've found myself tripping up over the idea of the location a lot, and I think online is much more clear word.


eldu (approved):
I think this is a lot clearer! 🎉

PR #266, opened 2023-08-15
build: Add lockfile back into git repo

RobertGemmaJr:
There seemed to have been issues with the lockfile before I started on the project but I generally think it's a non-issue? And we aren't able to do any depandabot checks without the lockfile

- Adds `package-lock.json` back into repo
- Upgrades `electron` to solve dependabot alert


hollandjg (approved):
Looks good to me! 
you might also be interested in https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups
…which makes for fewer PRs

PR #275, opened 2023-08-23
feat: Add CLI download script

### LINKED ISSUE #172: New download script
- [ ] Refactor to CJS Module
- [ ] Use regex check against `experimentID` instead of `sessionNumber`
- [ ] Giving `studyID` and `participantID` will download ALL sessions for that user
- [ ] File name can be individually indexed (e.g. `[studyID]_[participantI]D_[session].json`
- [ ] Prompt warning when overwriting files
- [ ] Need more compatible file names (ISO date uses `:` which is invalid on Windows)

```bash
npm run firebase:download -- studyID participantID [sessionID match] [outputRoot]
```

Multiple sessions should be downloaded to their own file, but the Nassa

**Date Range**

The Nassar lab is also looking for the ability to download across a date range for multiple participants on a given study. JS can only get CLI arguments by index so it will be pretty knarly to get them both working in the same script. Perhaps we look to have multiple download scripts?

```bash
npm run firebase:download -- studyID startDate [endDate] [outputRoot]
```

If just `startDate` we get all sessions since that date.

### LINKED ISSUE #250: Use `Commander` to re-write command line scripts
`Commander` provides much better command line functionalities, including documentation

### LINKED ISSUE #264: Use `Commander` for `firebase-download-data` script

RobertGemmaJr:
- Downloading data is now available through [inquirerer](https://github.com/SBoudrias/Inquirer.js)
- Script now has the ability to download multiple experiments of a given `studyID` and `participantID` at the same time!

Would love some opinions on the scaffolding of the switch statements and general organization here. It generally seems common to use a single `cli` script for the package. This will eventually include the creation of new Honeycomb tasks down the line, meaning we need the scaffolding of `ACTION` and `DEPLOYMENT` to be more than just switches. I had separated the `switch` cases into separate functions but I found that to be super repetitive and contained super shallow functions (e.g. `dowloadData` function just contained a switch for `downloadDataFirebase`, `downloadDataLocal`functions) .

The deletion script is just boilerplate for now, I'll work on that in another PR and merge both at the same time!

image

closes #250 
closes #264


package.json
```diff
@@ -21,12 +21,14 @@
...
     "@jspsych/plugin-survey-multi-select": "^1.1.1",
     "bootstrap": "^5.2.0-beta1",
+    "commander": "^11.0.0",
     "electron-log": "^4.4.8",
     "electron-squirrel-startup": "^1.0.0",
     "event-marker": "git+https://github.com/brown-ccv/event-marker.git",
     "execa": "^5.0.0",
     "firebase": "^10.1.0",
     "fs-extra": "^8.1.0",
+    "inquirer": "^9.2.10",
```
RobertGemmaJr:
```suggestion
```

RobertGemmaJr:
Check out #276!

cli.mjs
```diff
@@ -0,0 +1,258 @@
...
+  const validateParticipantFirebase = async (input) => {
+    // subcollection is programmatically generated, if it doesn't exist then input must not be a valid participantID
+    const studyIDCollections = await getParticipantRef(STUDY_ID, input).listCollections();
+    return studyIDCollections.find((c) => c.id === DATA_COL) ? true : invalidMessage;
+  };
+
+  // TODO: Enable downloading all study data at once
+  return await input({
+    // message: "Select a participant (* selects all ):", // ? Do we need the stuff in parentheses?
+    message: "Select a participant: ",
```
hollandjg:
```suggestion
    message: "Select a participant:",

```

cli.mjs
```diff
@@ -0,0 +1,258 @@
...
+    // default: "*",
```
hollandjg: would it be possible here to show the participant IDs as options?

cli.mjs
```diff
@@ -0,0 +1,258 @@
...
+async function studyIDPrompt() {
+  const invalidMessage = "Please enter a valid study from your Firestore database";
+  const validateStudyFirebase = async (input) => {
+    // subcollection is programmatically generated, if it doesn't exist then input must not be a valid studyID
+    const studyIDCollections = await getStudyRef(input).listCollections();
+    return studyIDCollections.find((c) => c.id === PARTICIPANTS_COL) ? true : invalidMessage;
+  };
+
+  return await input({
+    message: "Select a study:",
```
hollandjg, line 187: Would it be possible here to show the study IDs as options?

hollandjg (approved):
🚀 It worked for me! I've a couple of questions, but nothing that should stop a merge. Very Nice.

cli.mjs
RobertGemmaJr: Not yet, though I would definitely like to do this. Right now the data is nested in Firebase in such a way that the `participantID` doesn't actually exist, it's just part of the path to the data. The V4 release will restructure the data in Firebase so this is possible though!

cli.mjs
RobertGemmaJr: Same issue as the `participantID` ^^

RobertGemmaJr:
Also includes the deletion script from #276 now!

RobertGemmaJr:
- Download script prompts user for where to save the data (defaults to `.`)
- Download script prompts user for overwrite permissions if the file being downloaded already exists

PR #276, opened 2023-08-23
feat: Add data deletion to the CLI script

### LINKED ISSUE #265: Add a `firebase-delete-data` script
https://stackoverflow.com/questions/66013255/how-to-recursively-delete-collection-in-firestore

From slack message:

Good morning! I am using honeycomb and hosting via firebase. I noticed that my firebase storage usage was surprisingly high (over 8GB). I have a few different pilots saved to the same database, but the bottleneck seems to be that that I have many more participants that start my task but do not complete it. Each participant has a document associated with their ID and nested documents/collections with their data. I would like to easily delete participants by providing their prolific ID numbers as an array to some function and have it recursively delete all data associated with the subject. From what I understand of firebase, I cannot simply delete the document with the participant ID; it will not actually recursively delete the data. Any advice or insight would be great!

Oh great, I’ll look into this. It would be a super helpful resource to provide, especially since prolific recommends [deleting subject data](https://researcher-help.prolific.co/hc/en-gb/articles/360009092394-Approvals-rejections-returns#heading-3) if consent is withdrawn. Thanks!!

RobertGemmaJr:
- `delete` action on the CLI script now deletes user data
- User is prompted to confirm deletion
- Choosing an experiment is now ordered most recent first

image
image

closes #265


galenwinsor (approved):
Looks good to me!

hollandjg (approved):
🚀 This worked for me too!

cli.mjs
```diff
@@ -1,4 +1,4 @@
-import { checkbox, input, select } from "@inquirer/prompts";
+import { checkbox, input, select, confirm } from "@inquirer/prompts";
```
broarr, line 1: I don't expect you to change anything, but I wanted to let you know the new hotness for cli development is [commander](https://www.npmjs.com/package/commander)

https://npmtrends.com/commander-vs-inquirer

cli.mjs
```diff
@@ -77,15 +77,15 @@ async function downloadDataFirebase() {
...
         `${STUDY_ID}/${PARTICIPANT_ID}/${experimentID}.json`.replaceAll(":", "_");
 
       // TODO 172: Check for overwriting file?
       try {
         fsExtra.outputJSONSync(outputFile, experimentData, { spaces: 2 });
-        console.log("Data saved successfully: ", outputFile);
+        console.log("Data saved successfully:", outputFile);
       } catch (error) {
-        console.error("There was an error saving ", outputFile);
+        console.error("There was an error saving", outputFile);
```
broarr: Why not use template strings? I'm genuinely curious

broarr (approved):
Looks great to me!

cli.mjs
RobertGemmaJr: Think I'm just in the habit of not using them when the variable is at the end of the string. Will change!

cli.mjs
RobertGemmaJr: Yeah, I'm thinking we do some combination of both? So `inquirer` for the CLI prompts but someone could run `node cli.mjs download` and `commander` would handle the arguments

cli.mjs
broarr: Adding a var as an argument sometimes prints objects better. The template strings coerce to string (which gives you a lot of `[Object object]`). If your var is a string use the template strings, if your var is something more complex it might make sense to keep this as is!

cli.mjs
broarr: huh, interesting! Let me know how it works!

cli.mjs
galenwinsor: I also tend to not use template strings, mostly because it saves some awkward keystrokes.

RobertGemmaJr: Looks like they're strings so I'll update it!

PR #303, opened 2023-10-20
fix: Fix CLI download to work with older versions of data stored in Firebase

### LINKED ISSUE #299: Don't overwrite 'results` field when trials subcollection doesn't exist
`trials` sub collection is only used on newer tasks. For old tasks all the data is found in `experimentData` - the backwards compatibility will be easy

RobertGemmaJr:
Older versions of Honeycomb didn't store experiment data in the nested "trials" subcollection. The CLI was overriding those results with an empty array. We now skip that step if the Firebase document doesn't have that subcollection

### DIFF
--- a/cli.mjs
+++ b/cli.mjs
@@ -65,13 +65,17 @@ async function downloadDataFirebase() {
     const experimentSnapshot = await experimentRef.get();
     const experimentData = experimentSnapshot.data();
 
-    // Merge the experiment's trials' data into a single array
-    const trialsRef = experimentRef.collection(TRIALS_COL);
-    const trialsSnapshot = await trialsRef.orderBy("trial_index").get();
-    const trialsData = trialsSnapshot.docs.map((trial) => trial.data());
-
-    // Add the trials' data to the experiment's data as "results" array
-    experimentData["results"] = trialsData;
+    // Check for nested trials subcollection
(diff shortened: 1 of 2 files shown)


RobertGemmaJr:
Note to self - I need to update the CLI discussion post with this code once it's merged

cli.mjs
```diff
@@ -65,20 +65,25 @@ async function downloadDataFirebase() {
...
+    const trialsSnapshot = await experimentRef.collection(TRIALS_COL).orderBy("trial_index").get();
+    if (!trialsSnapshot.empty) {
+      // Add the trials' data to the experiment's data as "results" array
+      const trialsData = trialsSnapshot.docs.map((trial) => trial.data());
+      experimentData["results"] = trialsData;
+    } else {
+      // Running in an older version of Honeycomb where "trials" subcollection doesn't exist
+      // All experiment data used to be stored inside the experiment document -
+      // In this case "results" is already inside the experimentData object
+    }
```
digicosmos86, line 78: Why an empty else clause here? Are you just using it for the comment?

digicosmos86 (approved):
Looks great to me!

cli.mjs
RobertGemmaJr: Yeah, I wanted to make sure I noted that there is an "else" for backwards compatibility, it just doesn't have to do anything

PR #310, opened 2023-11-03
feat (index.html): Add a content security policy

RobertGemmaJr:
Adds a content security policy to `index.html`. There's a lot of Electron documentation that points to this as a best practice


digicosmos86 (approved):
LGTM!

PR #311, opened 2023-11-03
ref: Electron forge config file

RobertGemmaJr:
- Adds `forge.config.js` configuration file
- Updates ecma version
- Adds `package` script
- `package:<windows|mac|linux>` now build off the package script
- Consolidates the icons in `assets/icons`


forge.config.js
```diff
@@ -0,0 +1,57 @@
...
+        overwrite: true,
+      },
+      platforms: ["darwin"],
+    },
+    {
+      // Windows Distribution
+      name: "@electron-forge/maker-squirrel",
+      config: {
+        // TODO: Honeycomb icon (hexagon on docs)
+        // TODO: This will break when we merge with main
```
eldu: Is this an issue for existing honeycomb users or is it just the icon so they'll get like a missing image?

eldu (commented):
Passing through ! 🏃 

There are quite a few TODO statements here which look they they could be resolved by adding in the honeycomb icon png. Which could probably be done in this pr

I do like moving the forge config into it's own file though ! 

Also the check for the macOS build isn't going through

package.json
```diff
@@ -89,9 +92,10 @@
     "build:firebase": "dotenv -e env/.env.firebase npm run build",
     "build:turk-prolific": "dotenv -e env/.env.turk-prolific npm run build",
     "rebuild": "electron-rebuild",
-    "package:windows": "react-scripts build && electron-forge make --arch x64 --targets @electron-forge/maker-squirrel",
-    "package:linux": "react-scripts build && electron-forge make --arch x64 --targets @electron-forge/maker-deb",
-    "package:mac": "react-scripts build && electron-forge make --arch x64 --targets @electron-forge/maker-dmg"
+    "package": "npm run build && electron-forge make",
+    "package:windows": "npm run package -- --arch x64 --targets @electron-forge/maker-squirrel",
+    "package:linux": "npm run package -- --arch x64 --targets @electron-forge/maker-deb",
+    "package:mac": "npm run package -- --arch x64 --targets @electron-forge/maker-dmg"
```
digicosmos86: Looks like this does not support ARM macs natively right? Is there an option that supports a `universal` build?

digicosmos86 (approved):
Looks great! I only have one question about native support for ARM macs.

forge.config.js
RobertGemmaJr: The URL is point to `main` and we're changing the path to that file so we'll have to update it when this PR gets into `main` (it's going into a different branch right now)

RobertGemmaJr: Oh wait I can totally update the URL here for now, no one's going to be packaging based on this branch

RobertGemmaJr:
> Also the check for the macOS build isn't going through

@digicosmos86 @eldu have you run into node gyp errors on Mac before? The build is failing on the `npm install` and I'm not sure why

package.json
RobertGemmaJr: Looks like there is a universal value now, I'll swap it in!

eldu:
> > Also the check for the macOS build isn't going through
> 
> @digicosmos86 @eldu have you run into node gyp errors on Mac before? The build is failing on the `npm install` and I'm not sure why

@RobertGemmaJr Perhaps upgrade node-gyp? It's at 10.0.1
https://github.com/nodejs/node-gyp

eldu:
Oh node-gyp is not a direct dependency, nvm.

PR #334, opened 2023-11-21
feat: Export getSearchParam and update login fields based on it

### LINKED ISSUE #199: Use URLSearchParams for `getQueryVariable`

RobertGemmaJr:
The login page is now automatically filled out if `studyID` and `participantID`  are in the URLSearchParams

- Renames `getQueryVariable` as `getSearchParam`
- `getSearchParam` is now exported and used to read `studyID` and `participantID` from the URLSearchParams
- Renames state variables inside `<Login />`
- Removes `generateWaitSet` function as it's unused

closes #199

### DIFF
--- a/src/App/App.jsx
+++ b/src/App/App.jsx
@@ -90,9 +90,6 @@ export default function App() {
         }
       } else if (config.USE_FIREBASE) {
         // Fill in login fields based on query parameters (may still be blank)
-        // const query = new URLSearchParams(window.location.search);
-        // const studyId = query.get("studyID");
-        // const participantId = query.get("participantID");
         const maybeStudyID = getSearchParam("studyID");
         const maybeParticipantID = getSearchParam("participantID");
         if (maybeStudyID != null) setStudyID(maybeStudyID);
--- a/src/App/App.jsx
+++ b/src/App/App.jsx
@@ -16,7 +16,7 @@ import Login from "./components/Login";
 import { addToFirebase, validateParticipant } from "./deployments/firebase";
 
 import { config, taskSettings, taskVersion, turkUniqueId } from "../config/main";
-import { getProlificId } from "../lib/utils";
+import { getProlificId, getSearchParam } from "../lib/utils";
 
 /**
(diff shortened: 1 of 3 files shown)


src/App/App.jsx
```diff
@@ -90,18 +90,16 @@ export default function App() {
...
         // Fill in login fields based on query parameters (may still be blank)
-        const query = new URLSearchParams(window.location.search);
-        const studyId = query.get("studyID");
-        const participantId = query.get("participantID");
-        if (studyId) setStudyID(studyId);
-        if (participantId) setParticipantID(participantId);
+        const maybeStudyID = getSearchParam("studyID");
+        const maybeParticipantID = getSearchParam("participantID");
+        if (maybeStudyID != null) setStudyID(maybeStudyID);
+        if (maybeParticipantID != null) setParticipantID(maybeParticipantID);
```
eldu: Strict inequality?

```suggestion
        if (maybeStudyID !== null) setStudyID(maybeStudyID);
        if (maybeParticipantID !== null) setParticipantID(maybeParticipantID);
```

src/App/components/Login.jsx
```diff
@@ -8,47 +8,57 @@ export default function Login({
...
+  // Update local participantID if it changes upstream
+  useEffect(() => {
+    setParticipantID(initialParticipantID);
+  }, [initialParticipantID]);
+
+  // Update local studyID if it changes upstream
+  useEffect(() => {
+    setStudyID(initialStudyID);
+  }, [initialStudyID]);
+
```
eldu, line 24: Why would it change upstream?

eldu (commented):
Some questions ~

src/App/components/Login.jsx
RobertGemmaJr: I query parameters weren't auto-filling without these `useEffect`. I think it's because the state in `App.jsx` wasn't being set on the first render. I don't imagine folks will change the search parameters but if they do it will change the login fields which I think is nice

PR #348, opened 2023-12-06
ref: Change default Firebase ruleset

### LINKED ISSUE #147: Allow creation of particpant IDs at runtime (Firestore rules)
Clinical Trials - participants are pre-registered, experiment participant/study id must match
Prolific Online - Study IDs are pre-registered but anyone can participate

Right now there are Firestore rules such that the participant and study IDs must match what's already in Firestore in order to run the experiment (clinical trials). How do we let the experiment build such that this rule can be made optional?

https://github.com/brown-ccv/honeycomb/pull/144#discussion_r1128056480

RobertGemmaJr:
This ruleset allows the creation of `participantID`s and `studyID`s at runtime when using firebase. This is needed for working with Prolific which the majority of Firebase users are using at this point. I figure I'd just change the default rather than walk each lab through the change.

Also fixes an issue with the Content-Security Policy that was preventing API access to Firebase.

_**This is a change for version 3.3 of Honeycomb, merging into  a holding branch**_

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -4,7 +4,11 @@ service cloud.firestore {
     // participantID must be in the registered_participants array in the registered_studies/{studyID} document
     match /participant_responses/{studyID}/participants/{participantID} {
       allow create, read: 
-      if participantID in get(/databases/$(database)/documents/registered_studies/$(studyID)).data.registered_participants;
+      // Allows any combination of studyID and participantID to be created in Firebase
+      if true
+      // Forces participantID and studyID to already be in Firebase to log in
+      // if participantID in get(/databases/$(database)/documents/registered_studies/$(studyID)).data.registered_participants;
+
 
       // experimentID must be in the data subcollection
       match /data/{experimentID} {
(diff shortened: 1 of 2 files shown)


galenwinsor (approved):

firestore.rules
```diff
@@ -4,7 +4,11 @@ service cloud.firestore {
     // participantID must be in the registered_participants array in the registered_studies/{studyID} document
     match /participant_responses/{studyID}/participants/{participantID} {
       allow create, read: 
-      if participantID in get(/databases/$(database)/documents/registered_studies/$(studyID)).data.registered_participants;
+      // Allows any combination of studyID and participantID to be created in Firebase
+      if true
+      // Forces participantID and studyID to already be in Firebase to log in
```
eldu, line 9: Is there a way for researchers to preregister the participantID and studyID into Firebase? Perhaps add a link to the instructions here? And update documentation?

Also, part of me kind of feels like there should be a stricter default than a looser one. And then the documentation should make it clearer. However, from the way this PR is framed it sounds like like most researchers have pivoted to prolific and this is somewhat archaic

eldu (approved):
Approving and I added in a comment !

firestore.rules
RobertGemmaJr: Yeah, the idea of pre-registering participants isn't really a thing for online studies. It's mostly a hold out from clinic based studies where the participants are registered way ahead of time. 

I'll add a comment for a discussion post on this though!

PR #349, opened 2023-12-06
style: Add import rule for order

### LINKED ISSUE #325: Add eslint rule for import order
XMA Portal has a good example of this rule

RobertGemmaJr:
closes Add eslint rule for import order #325


public/electron.js
```diff
@@ -1,16 +1,16 @@
 // TODO #151: Can't use ES7 import statements here?
 
 // Modules to control application life and create native browser window
-const { app, BrowserWindow, dialog } = require("electron");
 const path = require("path");
+const { app, BrowserWindow, dialog } = require("electron");
 const ipc = require("electron").ipcMain;
```
anna-murphy: Why is this separated into two different lines? Could it be written like this?

```suggestion
const { app, BrowserWindow, dialog, ipcMain: ipc } = require("electron");
```

anna-murphy (approved):
Looks good! Im surprised there aren't more changes from this?

public/electron.js
RobertGemmaJr: That worked! Thanks!

PR #351, opened 2023-12-08
fix: Fix PsiTurk release location

RobertGemmaJr:
- PsiTurk build is now uploaded to the GitHub release it belongs to
- The PsiTurk build is folder is zipped before being uploaded

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -89,7 +89,7 @@ jobs:
         run: |
           echo "name=$(cat package.json | jq -r '.name')" >> $GITHUB_OUTPUT
           echo "version=$(cat package.json | jq -r '.version')" >> $GITHUB_OUTPUT
-        shell: bash
+        # shell: bash
 
       # Package on PsiTurk
       - name: 📦 Create Psiturk Build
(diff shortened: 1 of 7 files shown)


RobertGemmaJr:
[This is my reminder to create a new release once this is merged to confirm the action works]

eldu (commented):
I did a whole bunch of reading on uploading artifacts 😵 Looks like you wanted to add the artifacts onto the tags !

Quick question:

So this action runs on 
```
# Runs the build and package scripts when a new tagged release is publishe, created or edited
# Note that this runs for both releases and pre-releases: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
on:
  release:
    types: [created, edited, published]
```

I assume when a release is created, there is a tag that is also created?

Does in line 179 `tag: ${{ github.ref }}` refer to the correct tag when the release is edited or published?

Commenting for now... with questions.

RobertGemmaJr:
> I assume when a release is created, there is a tag that is also created?
> 
> Does in line 179 `tag: ${{ github.ref }}` refer to the correct tag when the release is edited or published?

@eldu So, yes, our current instruction is for folks to create a new tag whenever a new release is created. `github.ref` refers to the tag of the release that triggered the workflow. It will always refer to the correct release although tbh I'm unsure what it will do when file artifacts are already present in the release. I think it will overwrite naming conflicts but I haven't full tested that

galenwinsor (approved):

eldu (approved):

PR #357, opened 2023-12-12
ref: Check `config` at trial level

### LINKED ISSUE #355: Conditional check for `photodiodeGhostSpot` and `pdSpotEncode` for should be at the task level
The functions shouldn't handle any of the checks against `USE_PHOTODIODE` (besides throwing an error?). The developer should be expected to do that check at the task level.

RobertGemmaJr:
## What's Changed

- The `config` checks are done at the trial level, not inside `photodiodeGhostBox` and `pdSpotEncode`
- Tasks that are functions are made constants wherever possible
- Tasks that must be function are renamed as `build___Trial`
- Tasks that are constants are renamed as `___Trial`
- Fixes the parameters to `pdSpotZEncode` when it's being called
- Removes `export default` from trial files

## `config` Responsibility

This is one that I'm not entire sure which is better. I've chosen to have the trials check the `config` object rather than internal lib functions - `photodiodeGhostBox` and `pdSpotEncode`.

Part of the purpose/magic of HC is that it takes away from the developers need to worry about which environment the code is running in, which is why I do think having the config checks hidden inside the functions may be better. But ultimately a lot of this code is changing soon (because the event markers and photodiode will be available in both the browser AND native app) and I think expecting the caller to chose what they do/don't want to execute is best. This code is moving in that direction.

### Variables vs Functions

JsPsych expects all trials as an object. There are certain instances where we need to build a trial and must use a function (e.g. when we pass `jsPsych` around. However, in most cases, it's simpler to leave as a const variable.

Note that some keys to the trial (e.g. `stimulus`, `prompt`, etc.) can be written as callback functions that return the required type! This closer matches the jsPsych documentation and generally seems to be best practice, rather than functions that create the entire trial object. This PR covers a lot of those cases but doesn't cover 100%. I'll keep making changes as I progress.

### Naming

1. I think the  `build___Trial` is a bit more idiomatic of the difference between the variables and functions. Code that is a function is NOT a trial, it's building the trial object
2. The `___Trial` variable naming is a bit pedantic but I think it better communicates exactly what the objects are in a way that pure JS is not the greatest at. I think it's a useful naming scheme, especially considering HC will be most lab members first time writing code

### Removal of `export default`

This is just to keep consistency between files. I'm not 100% sure if I want one trial per file but I'll think about that for another PR. For now it makes it so the trials are imported in the same way across all trials


src/timelines/honeycombBlock.js
```diff
@@ -4,8 +4,16 @@ import { config, eventCodes, taskSettings } from "../config/main";
...
+ * Builds the blocks of trials that form the core of the Honeycomb experiment
+ *
+ * 1) A fixation dot is shown at the center of the screen
+ * 2) The stimulus image is shown and the user is prompted to press the correct key
+ *
+ * Note that the block is conditionally rendered and repeated based on the task settings
+ */
 function buildHoneycombBlock(jsPsych) {
-  const { honeycomb: honeycombSettings } = taskSettings;
+  const honeycombSettings = taskSettings.honeycomb;
```
galenwinsor, line 16: Nitpicky but it might be easier to read if config variables are in caps?

src/timelines/honeycombBlock.js
```diff
@@ -35,7 +43,7 @@ function buildHoneycombBlock(jsPsych) {
     },
     // Conditionally flashes the photodiode when the trial first loads
     on_load: () => {
-      if (config.USE_PHOTODIODE) pdSpotEncode(eventCodes.honeycomb, 1, config);
+      if (config.USE_PHOTODIODE) pdSpotEncode(eventCodes.honeycomb);
```
galenwinsor, line 46: Hmm I guess that would make this look weird, so maybe never mind.

src/timelines/preamble.js
```diff
@@ -1,25 +1,21 @@
...
-import startCode from "../trials/startCode";
-import { showName, showWelcome } from "../trials/welcome";
+import { enterFullscreenTrial } from "../trials/fullscreen";
+import { holdUpMarkerTrial } from "../trials/holdUpMarker";
+import { startCodeTrial } from "../trials/startCode";
+import { welcomeTrial, nameTrial } from "../trials/welcome";
+
+/** Builds the blocks of trials needed to start and setup the experiment */
+function buildPreambleBlock() {
+  const timeline = [nameTrial, enterFullscreenTrial, welcomeTrial];
```
galenwinsor, line 10: I like this pattern a lot! You're basically moving things from the top level into functions, right?

RobertGemmaJr: Right. And there's a HUGE difference between code that's inside a function and code that's inside a trial's callback function. Hoping to be a little more clear about that.

src/timelines/honeycombBlock.js
RobertGemmaJr: Definitely on my nice to haves but I'm not sure where to draw the line. Maybe just `SETTINGS`, `LANGUAGE`, and `ENV`? At the very least I think `env` isa  much better name than `config`

PR #358, opened 2023-12-12
fix: Only run release workflow when the release is created

RobertGemmaJr:
Releases must be tagged and a tag is always associated with 1 specific commit. Editing the release doesn't change the code it's using which means the executables will always be the exact same. If a user ever wanted to change the commit of a release they would have to delete the release, delete the tag, and then re-create the tag and release.

This is why I feel safe making this change. I have it in my TODOs to write a discussion post on the process of deleting and re-creating a release to change the commit it's built on!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,10 +1,10 @@
 name: 🚀 Release
 
-# Runs the build and package scripts when a new tagged release is publishe, created or edited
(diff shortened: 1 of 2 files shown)


RobertGemmaJr:
It is possible to force change the commit a tag is associated with but the user will still have to create a new release in order for that change to take effect.

broarr (approved):

PR #375, opened 2023-12-13
ref: Refactor task settings

RobertGemmaJr:
## What's Changed

- `config.json` is renamed `settings.json`
- `taskSettings` export is named `SETTINGS`

### Name Change

The `config.json` file holds configuration settings specific to the task. I think it's better to name as `settings` since we have multiple configuration settings (language, environment variables, and this) - it's a bit more specific.

This is one move towards that name change, and since they're just re-exports of the json I thought all caps was in order. Hence `LANGUAGE` and `SETTINGS`.

### Removing Settings in JS

I decided to remove the cascading of trial settings from `main.js`. I do still want this cascading effect to take place but that situation is specific to task settings that are stored in Firebase. I think only one base/fallback settings in the main repo  is okay and, if so, the separate JSON files is the better place for them. 

## Reviewer Qs

- Do you think having the single point of truth for the trial settings as a json file is a good idea?


galenwinsor (commented):
I think this looks good! I'm not entirely sure what you mean by the "cascade" of settings. But I think having a single source of truth in a JSON is great, if you don't need to change the settings frequently.

hetd54 (approved):
Also a fan of your description above! Agree with Galen 👍

broarr (approved):
This looks good to me, though I'm not sure I have all the context required. Why are `SETTINGS` and `LANGUAGE` the caps and the rest of the imports from config `camelCase`?

RobertGemmaJr:
> Why are `SETTINGS` and `LANGUAGE` the caps and the rest of the imports from config `camelCase`?

@broarr As of right now `SETTINGS` and `LANGUAGE` are the only ones from JSON files which is why I made them caps. In the end, yes, I think everything from this file should be exported as call caps.

Some of the exports will be removed in v3.3 and I'd like to actually remove the `config` option (and export things like `USE_ELECTRON` individually). Definitely a work in progress!

RobertGemmaJr:
> I'm not entirely sure what you mean by the "cascade" of settings.

The JSON will be the starting source of truth but the settings will be able to be configured on a study/participant level in Firebase. It's not implemented yet though.

PR #378, opened 2024-01-10
ref: `photodiodeGhostBox` is always a part of the prompt

### LINKED ISSUE #372: `photodiodeGhostBox` should always be passed as `prompt`
The `stimulus` is whatever it needs to be, we send the ghost box as a separate parameter

- [ ] buidCameraStartTimeline
- [ ] buildCameraEndTrial
- [ ] startCodeTrial

RobertGemmaJr:
## What's Changed

The expected type of the `stimulus` parameter changes based on the trial type whereas the prompt is always expected as an html string.

- Removes photodiode from the `camera` trials
- `startCode` uses the photodiode correctly
- `welcome` uses the photodiode correctly

All of the trials are now correctly showing the photodiode ghost box and flashing accordingly, yay! The hope is to move the photodiode functionality into a custom extension so the individual trials don't have to worry about these calls but that won't be until a v4

closes #372

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -141,19 +141,19 @@
       }
     },
     "node_modules/@babel/core": {
-      "version": "7.23.6",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.23.6.tgz",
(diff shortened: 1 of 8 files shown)


RobertGemmaJr:
I also found a bug with the video trial which has been fixed (the ID's were mismatched)

galenwinsor (approved):
This looks good to me! Not that I understand fully, but the changes seem fairly minor anyway.

[birdbox](https://app.graphite.dev/macros/brown-ccv)

broarr (approved):
LGTM!

PR #379, opened 2024-01-10
fix: Fix sending of event codes to Electron

RobertGemmaJr:
- Passes data correctly across the context bridge
- Early exit from `handleEventSend` when running in `DEV_MODE` (prevents the dialog box from popping up every trial)


public/electron/main.js
```diff
@@ -113,7 +113,11 @@ function handleSetConfig(event, config) {
 /**
  * Receives the Honeycomb config settings and passes them to the CONFIG global in this file
  * @param {Event} event The Electron renderer event
- * @param {Object} config The current Honeycomb configuration
+ * @param {Object} trigger The metadata for the event code trigger
+ * @param {string} trigger.comName The COM name of the serial port
+ * @param {Object} trigger.eventCodes The list of possible event codes to be triggered
+ * @param {string} productID The name of the product connected to the serial port
+ * @param {string} vendorID The name of the vendor connected to the serial prot
```
broarr, line 120: Are you actually using typescript to check these annotations?

broarr (approved):
Looks good!

public/electron/main.js
RobertGemmaJr: No, everything is pure JS

PR #405, opened 2024-02-01
build: Update Electron dependency and fix writeable stream

### LINKED ISSUE #399: Electron: Need to add closing `]}` to the JSON file if it's closed before the task completes
Has to do with the `app.on("before-quit"` logic

RobertGemmaJr:
- Updates `electron` to latest version
- Fixes an issue with the writeable stream by refactoring to `fs.appendFileSync`

Previously the last trial wasn't being written to correctly because the stream was "immediately" being used to append the closing `]}`. This is done synchronously down so it fixes the issue. I also cleanup up some of the logic for when a user quits the application before completing the experiment

closes #399

### DIFF
--- a/public/electron/main.js
+++ b/public/electron/main.js
@@ -27,8 +27,6 @@ let CONFIG; // Honeycomb configuration object
 let DEV_MODE; // Whether or not the application is running in dev mode
 
 let TEMP_FILE; // Path to the temporary output file
-// TODO: Delete, save temp file path?
-// let WRITE_STREAM; // Writeable file stream for the data (in the user's appData folder)
 let OUT_PATH; // Path to the final output folder (on the Desktop)
 let OUT_FILE; // Name of the final output file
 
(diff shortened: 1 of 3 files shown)


public/electron/main.js
```diff
@@ -141,6 +149,11 @@ function handleCheckSerialPort() {
   setUpPort().then(() => handleEventSend(TRIGGER_CODES.eventCodes.test_connect));
 }
 
+/**
+ * Sends the event_codes to the trigger port
+ * @param {} event The serial port event
+ * @param {number} code The event code to be recorder
```
tdivoll:
```suggestion
 * @param {number} code The event code to be recorded
```
I'm guessing the intended word is recorded?

public/electron/main.js
```diff
@@ -160,33 +173,38 @@ function handlePhotodiodeTrigger(event, code) {
 function handleOnDataUpdate(event, data) {
   const { participant_id, study_id, start_date, trial_index } = data;
 
-  // TODO @brown-ccv #397: Initialize writeable stream on login, not here
-  // Set the output file names/paths and initialize a writeable stream in the user's appData folder
-  if (!WRITE_STREAM) {
+  // Set the output path and file name if they haven't been
```
tdivoll:
```suggestion
  // Set the output path and file name if they are not set yet
```

tdivoll (commented):
Nice! Just little text cleanups from me.

PR #441, opened 2024-03-21
ci: Refactor Firebase actions to use the default scripts

### LINKED ISSUE #420: Node warnings for workflow actions
Update release actions to use node 20. If this works the general upgrade to node20 should be pretty simple?

### LINKED ISSUE #439: Firebase should use default deploy action
Change fro `npm install &&` to `npm ci &&` - that's the default that the Firebase CLI gets you to choose

RobertGemmaJr:
This is another PR for the 3.4 release.

Runs `npm ci && npm run build` instead of `npm install && npm run build`. It's safe to make the change because we now include the `package-lock.json` file


eldu (approved):

eldu (changes requested):
Actually wait, the checks don't pass?

eldu (approved):

PR #442, opened 2024-03-21
feat: Add prop validation to the JSX components

### LINKED ISSUE #223: Add prop validation to the JSX components
- [ ] App
- [ ] Error
- [ ] JsPsychExperiment
- [ ] Login

The `react/prop-types` Eslint rule can be removed and the prop types added from their

RobertGemmaJr:
- Remove `prop-validation` linter rule
- Adds prop validation to `Login` and `JsPsychExperiment` functions
- Refactos `App` components to pass props in the same order as the underlying component


package.json
```diff
@@ -33,6 +33,7 @@
     "fs-extra": "^11.1.1",
     "jspsych": "^7.2.3",
     "lodash": "^4.17.19",
+    "prop-types": "^15.8.1",
```
broarr, line 36: This is in place of typescript?

broarr (approved):

package.json
RobertGemmaJr: Right, I guess the react team moved prop types into its own library at some point. If/when Honeycomb becomes a separate library I think we'll move to typescript but I don't want to add the burden on new developers here

PR #447, opened 2024-03-26
chore: REMOVE IMAGE_INLINE_SIZE_LIMIT variable

RobertGemmaJr:
This is a stale environment variable that doesn't really matter? Images should be added to the `public/` folder and jsPsych now respects that. We should be dealing with any users adding images to `src/` anymore.

I know it makes for a few empty environment files but I think that's okay. I want to beef up the documentation of how to use the environment variables in which case have the "home" vs "clinic" is still pertinent


.nvmrc
```diff
@@ -1 +1 @@
-v18.17.1
```
eldu: This the sort of change I wouldn't expect in this PR. And instead in a PR that was about upgrading node.

But, I guess this is just some clean up after that fact?

eldu (approved):
Approving. I did add a comment, but nothing actionable really

.nvmrc
RobertGemmaJr: Hmm... I think the PR is out of date with `feat-3.4`, because I did have a  separate pr for upgrading node

PR #469, opened 2024-05-13
BREAKING CHANGE: Switch to `vite` with `electron-forge`

### LINKED ISSUE #419: Rename `package:*` scripts as `make:*` scripts
That's the name of the underlying `electron-forge` script. These mostly aren't run by the user (only the actions) so it won't be a  huge hurdle to get over

### LINKED ISSUE #421: Clean up React app with `electron-forge`
This should help? Just need to make sure everything continues to play nicely with the browser release

https://dev.to/mandiwise/electron-apps-made-easy-with-create-react-app-and-electron-forge-560e

### LINKED ISSUE #424: `electron-forge` should ONLY make the `build/` folder
Currently it's packaging everything in the repo, include the service account and things like that. May also need:

- `package.json`
- `package-lock.json`
- `node_modules/`

### LINKED ISSUE #464: HMR reload for electron-forge with vite
https://www.electronforge.io/config/plugins/vite#hot-module-replacement-hmr

RobertGemmaJr:
Switches Honeycomb to the Vite compiler. Note that we use  a Vite plugin (for `electron-forge` ) to run the app inside an Electron browser.

- Updates scripts to use `electron-forge` when running the application
- Removes `dev` script - we always pass a `.env.*` file
- Uses `import.meta.env` from Vite instead of `process.env`

## Firebase
- Removes `npm ci && npm run build:firebase` from the GH actions
- Adds `npm ci` and `npm run build:firebase` as predeploy scripts in `firebase.json`

Note that by doing this change the user doesn't have to worry about adding the predeploy script when they initialize Firebase hosting. One less step to simplify the onboarding process but I will have to update the documentation


RobertGemmaJr:
This is a pretty significant shift for Honeycomb would love as many opinions as possible! Happy to chat about it in person with folks if that's desirable

RobertGemmaJr:
The build times are at like 50% of what it used to take them with webpack 👀

broarr:
What does this break?

env/.env.clinic
```diff
@@ -1,3 +1,3 @@
-REACT_APP_USE_EEG="true"
-REACT_APP_USE_PHOTODIODE="true"
-REACT_APP_VOLUME="true"
\ No newline at end of file
+VITE_USE_EEG="true"
+VITE_USE_PHOTODIODE="true"
+VITE_VOLUME="true"
```
broarr, line 3: Did you need to change those?

broarr: I see, they're only used for the build

src/Electron/main.js
```diff
@@ -269,52 +291,63 @@ function handleSaveVideo(event, data) {
...
+  //     webPreferences: { preload: path.join(__dirname, "preload.js") },
+  //     fullscreen: true,
+  //     menuBarVisible: false,
+  //   });
+  //   // Hide the menu bar
+  //   mainWindow.setMenuBarVisibility(false);
+  // }
+  // // Load web contents at the given URL
+  // log.info("Loading URL: ", appURL);
+  // mainWindow.loadURL(appURL);
```
broarr: Do you need this?

broarr (approved):
Nothing is jumping out at me. Is there something you were particularly worried about?

RobertGemmaJr:
> What does this break?

Huh that's an interesting point. This is all part of the v4 release and I guess I kinda just internalized that the switch to a new compiler was a breaking change. But if all the logic is the same I suppose the whole point is that it doesn't break anything?

env/.env.clinic
RobertGemmaJr: Yeah, these are the "equipment" that users can turn on or off with environment variables

src/Electron/main.js
RobertGemmaJr: I clean this up in #473

broarr:
> Huh that's an interesting point. This is all part of the v4 release and I guess I kinda just internalized that the switch to a new compiler was a breaking change. But if all the logic is the same I suppose the whole point is that it doesn't break anything?

I think that's right, I looked for things that would break existing apis. I didn't see anything, but I could be wrong! Just wanted to check what you thought

PR #471, opened 2024-05-14
ref: `type: module`

RobertGemmaJr:
- Adds `type: module` to the package.json
- Specifies eslint as CJS
- Refactors `forge.config.js` and `.prettierrc.js` to be ES Modules
- Removes the `renderer.js` file, the `index.html` loads `index.jsx` directly
- Cleans up the vite config files a fair bit
- Specifies the vite builds as CJS

## Vite Builds
I've been able to dig into how the Electron forge plugin is working with the Vite configurations. Basically the `config` section of `"@electron-forge/plugin-vite"` the basic configuration for vite, which `vite.__.config.js` override. `forge.config.js`  is where we pass in the Electron files to be built and the specific vite configs handle the core of the building process.

The big takeaway is that, even though Vite handles the ESM format, **electron-forge builds the app as CJS**. All I really had to do was update the file extensions to specify CJS and vite runs the built files correctly. Obviously I would like the files to be built to ESM but that seems to be blocked by the `electron-forge` plugin itself.


broarr (approved):
Nice! I love modules

PR #474, opened 2024-05-14
build: Remove uneeded depencies

RobertGemmaJr:
I wanted to add the [electron-devtools-installer](https://github.com/MarshallOfSound/electron-devtools-installer) but it ended up being more work than it was worth.

I went through and cleaned up the deps we no longer need because of the change to vite!


broarr:
image
😱

broarr (approved):

RobertGemmaJr:
> image 😱

Turns out Babel is a pretty big dependency 😅

PR #475, opened 2024-05-14
fix: Add serialport code as its own file again

RobertGemmaJr:
Moves the serial port code back into its own file (`/lib/serialport.js`). 

Note that based on the current `forge.config.js` vite will build the serial port code into its own file. I could add it as a separate file in the config (and vite would build them separately) but I think  it's totally fine as is.


broarr (approved):
Net negative!

PR #476, opened 2024-05-15
BREAKING CHANGE: Update data paths

### LINKED ISSUE #417: 3.4.1: Rename task folder on the Desktop as `honeycomb-dev` when running the task in development mode
This way the data from testing in development mode is separated from the production

- [ ] `honeycomb-dev` folder on the Desktop
- [ ] `dev` subfolder in the temp path as well

### LINKED ISSUE #429: Use `app.getPath('temp')` for the temporary folder
3.4.1 release

Will have to update the documentation in regards to where this folder actually is

RobertGemmaJr:
- This changes the directory to which data is saved - any matlab scripts that labs may be using to parse the data will need to be updated

## Folder Structure

The big change here is how we're nesting the folders. The file follows the following structure:

/data/`mode`/`studyID`/`participantID`/`start_date`.json
- mode: `development` or `production`
- studyID: The ID of the study the participant logged in as
- participantID: The ID of the participant on the studyID the participant logged in as

That structure is nested inside [userData](https://www.electronjs.org/docs/latest/api/app#appgetpathname) while the participant is running through the experiment and then is copied into an output folder (Desktop/[getName](https://www.electronjs.org/docs/latest/api/app#appgetname)) when the participant finishes the experiment.

1. `data` is used to separate the output data from the experiment from any other data that may be included. I don't really have anything in mind but have noticed Honeycomb users tend to have other information (matlab scripts, papers, etc) in the folder. I like `data` as a better delineator here rather than jumping straight into `studyID`
2. `mode` is used to further delineate production data from data that was created during development. Most users have a separate lab computer for the production data so it's not much of a change but is definitely helpful for me. Plus we basically get it for free now that we're using Vite!

### User data
I was looking to move the "running" file save to the `temp` folder but everything I've been reading online points to saving the data in `userData` like it was. This way the data will stay there permanently. 

Note that the really important quirk of `userData` is the fact that is saves _partial_ run-throughs of the experiment. If a participant begins the experiment but quits the app before completing it (including app errors/crashes) the data will still be saved in `userdata` it just won't be copied over to the desktop. I think it's best to leave it as `userData` so that those partial run-throughs will be saved permanently, just hidden away.

## Output
Maintains that the data file(s) are created on the desktop at the end of the file. The file is copied from `userData`. Eventually I'd like for this path to be customizable but that's not a super pressing feature, the lab can change it with a single line of code if the really need

### DIFF
--- a/src/Electron/main.js
+++ b/src/Electron/main.js
@@ -37,20 +37,14 @@ const GIT_VERSION = JSON.parse(fs.readFileSync(path.resolve(__dirname, "version.
 // TODO @brown-ccv #436 : Use app.isPackaged() to determine if running in dev or prod
 // const ELECTRON_START_URL = process.env.ELECTRON_START_URL;
 const IS_DEV = import.meta.env.DEV;
-
-let CONFIG; // Honeycomb configuration object
 let CONTINUE_ANYWAY; // Whether to continue the experiment with no hardware connected (option is only available in dev mode)
-
-// let TEMP_FILE; // Path to the temporary output file
-// let OUT_PATH; // Path to the final output folder (on the Desktop)
-// let OUT_FILE; // Name of the final output file
-// let FILE_NAME; // The name of the data file
 
 const DATA_DIR = path.resolve(app.getPath("userData")); // Path to the apps data directory
 // TODO @brown-ccv: Is there a way to make this configurable without touching code?
 const OUT_DIR = path.resolve(app.getPath("desktop"), app.getName()); // Path to the final output folder (on the Desktop)
 let FILE_PATH; // Relative path to the data file. Becomes absolute when combined with DATA_DIR or OUT_DIR
 
+let CONFIG; // Honeycomb configuration object
 let TRIGGER_CODES; // Trigger codes and IDs for the EEG machine
 let TRIGGER_PORT; // Port that the EEG machine is talking through
 
--- a/src/Electron/main.js
+++ b/src/Electron/main.js
@@ -26,7 +26,6 @@ process.env["ELECTRON_DISABLE_SECURITY_WARNINGS"] = "true";
 
 /************ GLOBALS ***********/
 
-// These global variables are created by electron-forge
 /* global MAIN_WINDOW_VITE_DEV_SERVER_URL */
 /* global MAIN_WINDOW_VITE_NAME */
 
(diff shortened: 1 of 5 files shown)


broarr (approved):

src/Electron/main.js
```diff
@@ -177,46 +178,49 @@ function handlePhotodiodeTrigger(event, code) {
...
+    fs.appendFileSync(dataPath, `"start_time": "${start_date}",`);
+    fs.appendFileSync(dataPath, `"git_version": ${JSON.stringify(GIT_VERSION)},`);
+    fs.appendFileSync(dataPath, `"trials": [`);
   }
 
-  // Prepend comma for all trials except first
-  if (trial_index > 0) fs.appendFileSync(TEMP_FILE, ",");
+  const dataPath = getDataPath();
 
+  // TODO @RobertGemmaJr: Always write "proper" json (read json and append to it). Will need to update "before-quit" logic
```
RobertGemmaJr:
```suggestion
```

src/Electron/main.js
```diff
@@ -9,31 +9,29 @@ import _ from "lodash";
...
-// TODO: Handle version in renderer - pass into jspsych?
-// TODO: Just handle the commit id? I think that's probably fine
+// TODO: Preload function for passing this data into renderer - pass into jspsych?
+// TODO: Handle at runtime in a separate file not postinstall
 const GIT_VERSION = JSON.parse(fs.readFileSync(path.resolve(__dirname, "version.json")));
-const IS_DEV = import.meta.env.DEV && !app.isPackaged;
 
-let CONFIG; // Honeycomb configuration object
-let CONTINUE_ANYWAY; // Whether to continue the experiment with no hardware connected (option is only available in dev mode)
+// TODO @brown-ccv #436 : Use app.isPackaged() to determine if running in dev or prod
```
RobertGemmaJr:
```suggestion
```

src/Electron/main.js
```diff
@@ -226,47 +228,51 @@ function handleOnDataUpdate(event, data) {
...
+  fs.appendFileSync(dataPath, "]}");
+  log.info(`Finished writing experiment data to ${dataPath}`);
 
-  // Move temp file to the output location
-  const filePath = path.resolve(OUT_PATH, OUT_FILE);
   try {
-    fs.mkdirSync(OUT_PATH, { recursive: true });
-    fs.copyFileSync(TEMP_FILE, filePath);
-    log.info("Successfully saved experiment data to ", filePath);
+    // NEW
```
RobertGemmaJr:
```suggestion
```

PR #477, opened 2024-05-16
fix: Fix Package Workflow

RobertGemmaJr:
The Mac release of the package workflow is broken since we moved to the universal build of the app. This fixes the path names!


digicosmos86 (approved):
LGTM! I like the trash can emoji 🗑️

PR #479, opened 2024-05-16
feat: Add additional info to repo, data

### LINKED ISSUE #394: Write git sha to Firebase, remove deprecated information

RobertGemmaJr:
- Adds `PACKAGE_NAME` and `PACKAGE_VERSION` to `import.meta.env` in all processes
- Adds `PACKAGE_NAME` to jsPsych data object as `app_name`
- Adds `PACKAGE_VERSION` to jsPsych data object as `app_version`
- Removes `taskVersion` and `taskName` from `config.js`
- Re-orders the appearance of app related data in the jsPsych data object (important for `.csv` downloads) 

## Git Versioning
- Honeycomb now checks the git sha & ref at _runtime_ while in development
- The logic for handling the the git sha & ref in the built application is the same
- The git information is passed directly into the jsPsych data object as `app_commit`

Note that in order to get this to work I had to change the `useMemo` to be an async `useEffect` when building the jsPsych class. I'd definitely like to move that code to the `onSubmit` of the login form but I'll tackle that in another PR


src/App/deployments/firebase.js
```diff
@@ -67,8 +67,13 @@ export async function validateParticipant(studyID, participantID) {
 export async function initParticipant(studyID, participantID, startDate) {
   try {
     const experiment = getExperimentRef(studyID, participantID, startDate);
+
     await experiment.set({
       // TODO @brown-ccv #394: Don't handle any of this here? Let everything be done in jsPsych
```
YUUU23: I might've resolved a very minor merge conflict here with comments that resulted in some syntax errors; will get this changed back

YUUU23 (approved):
I agree with moving initialization of the `jsPsych` object for each experiment to `onSubmit` on the login form. If I am understanding the intention correctly, I agree it makes more sense to init the experiment when submit is clicked. 

Also, I was wondering about the idea if we could possibly have a `getSHA` and maybe `getRef` function to retrieve those values (would be used in `getCommit` function for `Electron/main.js` and `versions.js`) inside utilities functions so we can also use the same helper functions to retrieve and write `GIT SHA` to firebase?

PR #480, opened 2024-05-17
BREAKING CHANGE: Remove VOLUME equipment

### LINKED ISSUE #401: Remove `USE_VOLUME` environment variable
There's no real "Honeycomb logic" in regards to whether or not the task plays audio. If a user want's a task with audio they should just add that given trial

RobertGemmaJr:
- Removes the `VOLUME` environment variable

The environment variable used to be used to conditionally play a beep but now that logic is built out into its own trial, we don't really need it anymore. Not many labs use audio these days, if one does we can circle back to it as a special case for them

closes #401


broarr (approved):

broarr:
That definitely looks like a breaking change

YUUU23 (approved):

PR #481, opened 2024-05-17
ref: Remove Python

### LINKED ISSUE #278: Delete Python OR add `miniforge` to dependencies for Python management
Add to both `winget` and `brew`

RobertGemmaJr:
- Deletes Python as a dependency of the project 

Note that this was made possible because 1) We removed the PsiTurk deployment and 2) The upgrade to SerialPort 12 means the native node modules are built without Python. Begone `node-gyp` !!!

closes #278


eldu (approved):
AMAZING !!!!!!

RobertGemmaJr:
Marking as draft because there is still some issues going on here. I think we may need to keep it but we shall see!

PR #491, opened 2024-06-12
ref: Cleanup the login screen

RobertGemmaJr:
Most of this work was originally done as part of the Gearshift project, bringing it into Honeycomb.

@YUUU23 Definitely comment on any questions or reply in GitHub directly! If you want I'm happy to chat more about the setup of the project or this specific PR more if you'd like.

- Adds a state variable for `isLoading`. This updates the text of the button so the user knows the form is submitting under the hood
- Marks the input fields as `required` and removes the button's `disabled` check
    - By marking the field as `required` (more info [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required)) we utilize the HTML's [validation ](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation)instead of providing our own
    - We generally want the user to be able to press the button as often as possible! The form can always be _submitted_ it's just not always _valid_. Now we use the browser to do that initial check!
- Replaces the `type` field in the input components with `name`
    - This was done in error, it should have been the name the whole time
    - The valid input field types are listed [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types). Note that the default type is "text".
    - The `<Form.Group/>`'s `controlId` matches the `name` field of the input. Makes sure they're matched up!

- Updates the error message of the login screen to be more generic
   - While the error usually happens when someone logs in with the incorrect information it _actually_ errors anytime the project cannot connect to Firebase. I think this message is specific enough to the login information but prevents confusion when a user/developer definitely knows they're logging in with the correct information 
- Adds a link to the [Quick Start guide](https://brown-ccv.github.io/honeycomb-docs/docs/quick_start#2-add-a-file-for-the-task) inside the experiment's index file.

## A note on Branches

Note how this PR is merging into the branch `feat-v3.4.1`. Honeycomb is a template repository which adds a unique challenge when handling our releases. We must keep the `main` branch completely clean because that's the code that will get copied whenever a user creates a new Honeycomb project.

We get around this issue by having these "holding" branches for each version number. Once all of the changes for v3.4.1 are completed we'll have one final PR to bring the changes to main and then generate a new release!


broarr (approved):

YUUU23 (approved):
@RobertGemmaJr 

Thank you for the detail descriptions -- they really helped me to understand the changes on this PR well!

src/App/components/Login.jsx
```diff
@@ -11,23 +11,31 @@ export default function Login({
   // State variables for login screen
   const [participantID, setParticipantID] = React.useState(initialParticipantID);
   const [studyID, setStudyID] = React.useState(initialStudyID);
+
+  // State variable for handling errors
   const [isError, setIsError] = React.useState(false);
 
+  // State variable for handling loading states
```
RobertGemmaJr, line 18: Hey! Spelling error
```suggestion
  // State variable used to loading states
```

PR #495, opened 2024-06-14
fix: ci for mac

eldu:
https://www.electronforge.io/guides/code-signing/code-signing-macos#osxsign-options


RobertGemmaJr:
I'm going to merge this back in, thanks so much @eldu!

forge.config.js
```diff
@@ -29,6 +30,7 @@ export default {
       config: {
         options: {
           icon: "assets/icons/icon.png",
+          format: "ULFO",
```
eldu, line 33:
```suggestion
```

eldu: @RobertGemmaJr we can delete this line! I added this in but it doesn't seem to do anything

eldu:
```suggestion
```

PR #496, opened 2024-06-19
feat: add sample countdown trial that counts down before the start of…

### LINKED ISSUE #212: Add `Countdown` example trial
Will be a `NO_KEYS` trial, default to count down from 3-0

YUUU23:
- add example `countdown` trial in `trials` folder inside honeycomb for counting down given millisecond (ms) before another trial begins: 
  - in `src/experiment/trials/countdown.js`: 
       - helper functions to convert ms to minutes and seconds to format a time string (00:00) 
       - `buildCountdownTrial` function to return the actual JS trial object that counts down and updates the time string every few ms until counted down to 0 
   - in `src/experiment/honeycomb.js`: an example usage of the `countdown` trial to countdown 6000ms and inserted in the experiment timeline

### DIFF
--- a/src/experiment/honeycomb.js
+++ b/src/experiment/honeycomb.js
@@ -3,6 +3,7 @@ import { buildHoneycombProcedure } from "./procedures/honeycombProcedure";
 import { buildStartProcedure } from "./procedures/startProcedure";
 
 import { buildDebriefTrial, instructionsTrial, preloadTrial } from "./trials/honeycombTrials";
+import { buildCountdownTrial } from "./trials/countdown";
 
 /**
  * ! This file should not be edited! Instead, create a new file with the name of your task
@@ -47,8 +48,12 @@ export function buildHoneycombTimeline(jsPsych) {
   // Builds the trials that make up the end procedure
   const endProcedure = buildEndProcedure(jsPsych);
 
+  // Builds a countdown trial that counts down for 6000ms
+  const countdownTrial = buildCountdownTrial(6000);
+
   const timeline = [
     startProcedure,
+    countdownTrial,
     preloadTrial,
     instructionsTrial,
     honeycombProcedure,
--- a/src/experiment/honeycomb.js
+++ b/src/experiment/honeycomb.js
@@ -48,8 +48,8 @@ export function buildHoneycombTimeline(jsPsych) {
   // Builds the trials that make up the end procedure
   const endProcedure = buildEndProcedure(jsPsych);
 
-  // Builds a countdown trial that counts down for 6000ms
-  const countdownTrial = buildCountdownTrial(6000);
+  // Builds a countdown trial that counts down for 3000ms
+  const countdownTrial = buildCountdownTrial(3000);
 
   const timeline = [
     startProcedure,
--- a/src/experiment/trials/countdown.js
+++ b/src/experiment/trials/countdown.js
@@ -4,7 +4,7 @@ import { h1 } from "../../lib/markup/tags";
 /**
  * Returns millisecond to minute
  *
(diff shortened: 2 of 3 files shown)


src/experiment/trials/countdown.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/**
+ * a sample countdown trial that counts down ms before another trial begins
+ *
+ * @param {*} ms - millisecond to countdown
+ * @returns a JS object as a trial
+ */
+export function buildCountdownTrial(ms) {
+  const waitTime = ms;
```
RobertGemmaJr: I would rename the parameter here to be `waitTime` instead of `ms`! You note that the time is in milliseconds in the docstring and that's the most common way time is handled in JS so you can be a little more explicit with the variable name of the function

src/experiment/trials/countdown.js
```diff
@@ -0,0 +1,64 @@
+import htmlKeyboardResponse from "@jspsych/plugin-html-keyboard-response";
+import { h1 } from "../../lib/markup/tags";
+
+/**
+ * Returns millisecond to minute
+ *
+ * @param {*} ms - millisecond
```
RobertGemmaJr: Can you update the types here? It looks like they're almost always a `number`

```suggestion
 * @param {number} ms - millisecond
```

src/experiment/honeycomb.js
```diff
@@ -47,8 +48,12 @@ export function buildHoneycombTimeline(jsPsych) {
   // Builds the trials that make up the end procedure
   const endProcedure = buildEndProcedure(jsPsych);
 
+  // Builds a countdown trial that counts down for 6000ms
+  const countdownTrial = buildCountdownTrial(6000);
```
RobertGemmaJr: Let's make this 3000? Just so it's not taking up too much time

RobertGemmaJr (changes requested):
Let's chat about this a little closer on Tuesday? We have a special utils file in the lib/ folder that I think some of these functions would work perfectly! Love the way they're broken out and a little renaming would make them perfect for the entire project!

Fantastic work 🙌

YUUU23:
@RobertGemmaJr Thank you for these feedbacks! I will go ahead and make those changes, and let's talk about it more on Tuesday's meeting! : )

src/experiment/trials/countdown.js
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Returns millisecond to minute
+ *
+ * @param {number} ms - millisecond
+ * @returns minute value
+ */
+function getMinute(ms) {
+  return Math.floor(ms / 1000 / 60);
+}
```
RobertGemmaJr: This can be moved into `src/lib/utils`!

src/experiment/trials/countdown.js
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Given total minute and total millisecond, return the seconds
+ *
+ * @param {number} ms - millisecond
+ * @param {number} min - minute
+ * @returns the seconds in number
+ */
+function getSeconds(ms, min) {
+  return Math.floor((ms - min * 1000 * 60) / 1000);
+}
```
RobertGemmaJr: This can be moved into `sec/lib/utils`!

YUUU23 (commented):
All changes discussed about should be changed!

RobertGemmaJr (approved):
Wonderful! Thank you for this!

RobertGemmaJr:
@YUUU23 You can actually merge this pull request into `3.4.1`! Sorry if that was confusing

PR #502, opened 2024-06-26
refactor: full screen trial with additional data from gearshift proje…

### LINKED ISSUE #486: Add additional data to the full screen trial
Function is defined in Gearshift, need to bring into Honeycomb.

- width
- height
- pixel device ratio

YUUU23:
change full screen trial with code from Gearshift that has additional data about user's screen size


RobertGemmaJr (approved):
Awesome!

PR #505, opened 2024-07-06
feat: add getCurrentTrialData form gearshift into utils.js

YUUU23:
from gearshift, add to `src/libs/util.js` `getCurrentTrialData` function 
(please let me know if this function should be moved somewhere else! I thought it made the most sense to be added as a utilities function)

PR #506, opened 2024-07-06
feat: add getLastTrialData from gearshift into src/lib/utils.js

### LINKED ISSUE #499: Add `getLastTrialDataUtility` function from GearShift
```js
/**
 * Retrieves the data object from the last trial in JsPsych
 * @param {JsPsych} jsPsych jsPsych instance being used to run the task
 * @returns {Object}
 */
export function getLastTrialData(jsPsych) {
  const dataCollection = jsPsych.data.getLastTrialData();
  return dataCollection.trials[0];
}
```

YUUU23:
from gearshift, add to `src/libs/util.js` getLastTrialData function 
(should this function go here also similar to `getCurrentTrialData` from this [PR](https://github.com/brown-ccv/honeycomb/pull/505)?)

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -111,3 +111,13 @@ export function getSeconds(ms) {
 export function getTimeString(ms) {
   return `${getMinute(ms)}:${getSeconds(ms).toString().padStart(2, "0")}`;
 }
+
+/**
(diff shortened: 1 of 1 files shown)


RobertGemmaJr (approved):
Yup this is great! The perfect file for it as well 😄

PR #507, opened 2024-07-06
Add get random element

### LINKED ISSUE #485: Add `getRandomElement` utility function
Function is defined in Gearshift, need to bring into Honeycomb

YUUU23:
I found the `getRandomElement` function in this repo that seems like a version of gearshift [repo](https://github.com/brown-ccv/jehli-gearshift-task/blob/6844fe4d837c01d06a19f77f572027a72965f27e/src/lib/utils.js#L97)

I'm guessing that this utilities function is still the same; please let me know if anything needs to be changed or if this function should be moved somewhere else : ) 

Also changed the general `object` type to `JsPsych` since we will define that in `typedef.js`


RobertGemmaJr (approved):
Love it!! 🚀

PR #509, opened 2024-07-12
refactor: holdUpMarker as basic trial to line up photodiode spot and …

### LINKED ISSUE #466: Refactor `holdUpMarker` as a basic trial for lining up the photodiode spot
Cleaned up trial is here:

https://github.com/rossan750/nback_honeycomb/blob/main/src/trials/holdUpMarker.js

YUUU23:
* Move `holdUpMarker` code in to change original to a basic trial 
* Move `countdownTrial` to before `preloadTrial` in the timeline so the countdown will begin after serialport pop-up box (all set-up completed)

### DIFF
--- a/src/experiment/trials/holdUpMarker.js
+++ b/src/experiment/trials/holdUpMarker.js
@@ -1,32 +1,13 @@
 import htmlButtonResponse from "@jspsych/plugin-html-button-response";
+import { LANGUAGE } from "../../config/main";
+import { h1 } from "../../lib/markup/tags";
+import { photodiodeGhostBox } from "../../lib/markup/photodiode";
 
-import { config, LANGUAGE } from "../../config/main";
-import { eventCodes } from "../../config/trigger";
-import { pdSpotEncode, photodiodeGhostBox } from "../../lib/markup/photodiode";
-import { div, h1, p } from "../../lib/markup/tags";
-
-// TODO @brown-ccv #330: Custom extension for EEG - this is initializeTriggerBox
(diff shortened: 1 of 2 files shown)


src/experiment/trials/holdUpMarker.js
```diff
@@ -1,32 +1,13 @@
...
-  prompt: function () {
-    let holdUpMarkerPrompt = p(LANGUAGE.trials.holdUpMarker);
-
-    // Conditionally add the photodiodeGhostBox
-    if (config.USE_PHOTODIODE) holdUpMarkerPrompt += photodiodeGhostBox;
-
-    return holdUpMarkerPrompt;
-  },
+  stimulus:
+    h1(LANGUAGE.trials.eventMarker.connected, { style: "color: green;" }) +
```
RobertGemmaJr: Let's remove this line for now. We don't actually check whether or not the event marker is connected so it doesn't make much sense to handle it here

```suggestion
```

RobertGemmaJr (commented):
I'm now realizing it's probably best to handle this as version 3.4.2 of the project. With the current setup the serial port is not actually connected until the following trial.

I think it would be best to handle some sort of [call-function](https://www.jspsych.org/v7/plugins/call-function/) trial that calls the Electron `setupPort` function before continuing. This will take a little while, and tbh I think having a really solid setup for the serial port connection is worthy of it's own patch version - it's something that's always kind of been an issue with Honeycomb.

RobertGemmaJr:
Let's make the minor change I mentioned and then we'll create a 3.4.2 holding branch. From there we'll add the new trial as its own PR!

PR #510, opened 2024-07-12
Add mockbind

### LINKED ISSUE #460: Serial point connection with `MockBindings` when running in `CONTUNUE_ANYWAY` mode
https://serialport.io/docs/api-binding-mock

YUUU23:
- Add mock binding when user clicks continue anyway 
- Change `log.info` to not include port 
- `CONTINUE_ANYWAY` throws warning that it's never used after deleting check on if continue anyway is true, we return from `handleEventSend()`; I'm not sure if we can remove this variable entirely? Otherwise to resolve this, I added a check for that if it's `CONTINUE_ANYWAY` is false, we then set it to true

### DIFF
--- a/public/electron/main.js
+++ b/public/electron/main.js
@@ -8,6 +8,8 @@ const { app, BrowserWindow, ipcMain, dialog } = require("electron");
 const log = require("electron-log");
 const _ = require("lodash");
 
+const { MockBinding } = require("@serialport/binding-mock");
+const { SerialPortStream } = require("@serialport/stream");
 const { getPort, sendToPort } = require("./serialPort");
 
 // TODO @brown-ccv #460: Add serialport's MockBinding for the "Continue Anyway": https://serialport.io/docs/guide-testing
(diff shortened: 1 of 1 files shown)


public/electron/main.js
```diff
@@ -416,7 +415,21 @@ function handleEventSend(code) {
         break;
       case 2:
         // User selects "Continue Anyway", we must be in dev mode
-        CONTINUE_ANYWAY = true;
+        // resolve variable unused warning
+        if (CONTINUE_ANYWAY == false) {
+          CONTINUE_ANYWAY = true;
+        }
```
eldu: I believe this can be simplified without the if statement

```suggestion
        CONTINUE_ANYWAY = true;
```

eldu: Is this mainly a statement to use the `CONTINUE_ANYWAY` variable?

YUUU23: I agree this is very redundant -- yes, it is primarily to use this variable to pass the eslint tests. I was wondering since it gives this unused variable warning, would it be okay to just delete this variable overall?

RobertGemmaJr: We can safely remove the variable! That variable was only there to skip sending data to a port that didn't exist, but now we have a mock port to send it to

RobertGemmaJr (changes requested):
This looks fantastic! Adding the mock port has been a much needed addition for quite some time now. Thank you for tackling it!

You can safely remove the `CONTINUE_ANYWAY` variable, then this is ready to go

YUUU23:
> This looks fantastic! Adding the mock port has been a much needed addition for quite some time now. Thank you for tackling it!
> 
> You can safely remove the `CONTINUE_ANYWAY` variable, then this is ready to go

This variable should be removed!

RobertGemmaJr (approved):
Wonderful!

PR #512, opened 2024-07-18
rm bootstrap

### LINKED ISSUE #344: Remove `bootstrap` dependency
Will make the upgrade to React 18 much easier and it's not really necessary - everything can be handled with pure CSS

- [ ] `Login.jsx`
- [ ] `Error.jsx`
- [ ] `camera.js` (`buildCameraStartTrial` and `buildCameraEndTrial`)

YUUU23:
In `Login.jsx`: 
- refactor form and button to use vanilla HTML + CSS (styling may not be exactly like bootstrap styling, but should be very similar); all new CSS added in place of the bootstrap styling is in `index.css`
- remove all related bootstrap imports (also removed in `App.jsx`)

In `Camera.js`; 
- remove centering camera bootstrap classes and added CSS to simply center those elements when setting up and ending camera

Change `alert alert-danger` to `alert-danger` CSS class with similar styling as bootstrap's alert-danger class

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,6 @@
         "@jspsych/plugin-instructions": "^1.1.3",
         "@jspsych/plugin-preload": "^1.1.2",
         "@jspsych/plugin-survey": "^1.0.1",
-        "bootstrap": "^5.2.0-beta1",
         "electron-log": "^5.0.0",
         "electron-squirrel-startup": "^1.0.0",
         "execa": "^8.0.1",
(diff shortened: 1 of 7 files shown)


YUUU23:
Should I remove the boostrap dependency manually from `package.json` or does it get automatically removed when rebuilding we do not use bootstrap anywhere?

RobertGemmaJr:
> Should I remove the boostrap dependency manually from `package.json` or does it get automatically removed when rebuilding we do not use bootstrap anywhere?

You'll have to remove it manually! You can run `npm uninstall bootstrap react-bootstrap` to uninstall the dependencies

RobertGemmaJr (commented):
Fantastic! Just uninstall the dependencies we no longer need and this will be good to go!

YUUU23:
> Fantastic! Just uninstall the dependencies we no longer need and this will be good to go!

Just made this change! Would you be able to check one more time if everything looks good? 
Thanks!

RobertGemmaJr (approved):
Looks fantastic!

PR #513, opened 2024-07-18
add vertical spacing to form element after removing bootstrap (vanill…

### LINKED ISSUE #195: Add vertical spacing between form elements
Note that Bootstrap is going to be removed so it's best to use something other than that #344

YUUU23:
add vertical spacing to the form elements after refactoring them to vanilla css instead of bootstrap

### DIFF
--- a/src/App/index.css
+++ b/src/App/index.css
@@ -45,7 +45,7 @@ body,
 
 .form-input {
   width: 100%;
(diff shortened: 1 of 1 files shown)


RobertGemmaJr (approved):
LGTM! 🚀

PR #515, opened 2024-07-26
add: callFunction trial to call setUpPort after holdupMarkerTrial

### LINKED ISSUE #466: Refactor `holdUpMarker` as a basic trial for lining up the photodiode spot

YUUU23:
- `callFunction.js`: JSPsych trial that calls function passed in
- `startProcedure.js`: calls buildCallFunctionTrial in `callFunction.js` passing in `checkSerialPort()` that access the electron process to call `setUpPort()` after `holdUpMarkerTrial`


package.json
```diff
@@ -26,6 +26,7 @@
     "@electron/fuses": "^1.8.0",
     "@fortawesome/fontawesome-free": "^6.4.2",
     "@jspsych/plugin-audio-keyboard-response": "^1.1.3",
+    "@jspsych/plugin-call-function": "^2.0.0",
```
RobertGemmaJr: Hey! Can you actually set this to be version ^1.1.0? jsPsych just released a new version and all of their plugins have been updated as well, we want the one that's major version #1
```suggestion
    "@jspsych/plugin-call-function": "^2.0.0",
```

src/experiment/trials/callFunction.js
RobertGemmaJr, line 1: Awesome! I love the way you added it as a parameter

src/experiment/trials/callFunction.js
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @param {function} func will call this function
+ * @returns JS object as a trial
+ */
+export function buildCallFunctionTrial(func) {
+  return {
+    type: callFunction,
+    func: () => {
+      func;
+    },
```
RobertGemmaJr: Little oddness of JavaScript here! Because we're passing a callback function to the property `func` we actually need to supply a callback function, and not just the _call_ to the function itself. There's a  couple ways that we can do this:


```suggestion
    // 1) This way is best
    func: func,
    // 2) Or this would work too
    func: () => {
      func()
    },
```

Notice how with option 2 we cannot pass any parameters to the function - even if we define them in `startProcedure.js`! Because of this I would go with option 1, and then in `startProcedure` you pass a callback function to the code we actually want to execute:

```js
procedure.push(buildCallFunctionTrial(() => { window.electronAPI.checkSerialPort() }));
```

RobertGemmaJr (changes requested):
Looks fantastic! The structure of the code is excellent. Just  a minor thing about how JS executes the callback functions and you'll be on your way!

src/experiment/trials/callFunction.js
YUUU23: Thank you!!

YUUU23:
Please see this PR https://github.com/brown-ccv/honeycomb/pull/524 for the change of merging it into 3.4.2 instead!

YUUU23:
Same code merged here https://github.com/brown-ccv/honeycomb/pull/524

PR #516, opened 2024-07-27
add: commander to optionally provide delete or download along with st…

### LINKED ISSUE #289: Add `commander` to CLI script
- `npm run cli [download|delete]` skips the first question
- `npm run cli [download|delete] [studyID]` skips the first two questions
- `npm run cli [download|delete] [studyID] [participantID]` skips the first three questions

YUUU23:
- import `commander` and initialize commander to be used 
 - when no command provided, default action is to run `main()` as before, all prompts will be asked
    - checks if the studyID / participantID passed in through argument are valid, return from `main()` if not (validating functions originally in prompt functions are pulled out for this purpose) 
 - when user runs `npm run cli download [study ID] [participant ID]` or `npm run cli delete [study ID] [participant ID]`, relative global variables (`ACTION`, `STUDY_ID`, `PARTICIPANT_ID`) are set to what's provided through command line input, skipping those prompts since we have a value assigned to the global variables already 
    -  A message is also printed out to let user know which action and args are passed in through the command line
- Tested by interacting with the command line with different combinations  

_I accidentally pushed the code to the actual `feat-v4` branch, but I reverted this commit (now in this PR)! Sorry about this_


cli.mjs
```diff
@@ -18,16 +19,73 @@ let OUTPUT_ROOT; // The root in which data is saved
...
+    ACTION = "delete";
+    STUDY_ID = studyID;
+    PARTICIPANT_ID = participantID;
+  });
+commander.parse();
+
+// print message if download or delete provided, along with optional args provided
+if (ACTION != undefined) {
+  console.log(
+    `${ACTION} data from Firebase given ${STUDY_ID == undefined ? "" : `study ID: ${STUDY_ID}`} ${PARTICIPANT_ID == undefined ? "" : `and participant ID: ${PARTICIPANT_ID}`}`
```
RobertGemmaJr: Can you make these strict equality? `===`

cli.mjs
```diff
@@ -18,16 +19,73 @@ let OUTPUT_ROOT; // The root in which data is saved
...
+  );
+}
+
 /** -------------------- MAIN -------------------- */
 
 // TODO @brown-ccv #289: Pass CLI arguments with commander (especially for action)
 async function main() {
```
RobertGemmaJr, line 62: Same here! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

cli.mjs
```diff
@@ -18,16 +19,73 @@ let OUTPUT_ROOT; // The root in which data is saved
...
   // TODO @brown-ccv #291: Enable downloading all study data at once
-  STUDY_ID = await studyIDPrompt();
+  if (STUDY_ID == undefined) {
+    STUDY_ID = await studyIDPrompt();
+  } else {
+    // when args directly passed in through CLI, check if study is valid
+    const hasStudy = await validateStudyFirebase(STUDY_ID);
+    if (hasStudy != true) {
+      console.error(hasStudy);
+      return;
```
RobertGemmaJr: What happens on this return? Are there any more prompts or does the program just end?

cli.mjs
```diff
@@ -185,19 +242,19 @@ async function deploymentPrompt() {
...
+const validateStudyFirebase = async (input) => {
   const invalidMessage = "Please enter a valid study from your Firestore database";
-  const validateStudyFirebase = async (input) => {
-    // subcollection is programmatically generated, if it doesn't exist then input must not be a valid studyID
-    const studyIDCollections = await getStudyRef(input).listCollections();
-    return studyIDCollections.find((c) => c.id === PARTICIPANTS_COL) ? true : invalidMessage;
-  };
+  // subcollection is programmatically generated, if it doesn't exist then input must not be a valid studyID
+  const studyIDCollections = await getStudyRef(input).listCollections();
+  return studyIDCollections.find((c) => c.id === PARTICIPANTS_COL) ? true : invalidMessage;
```
RobertGemmaJr: Hmm... I know this is code I wrote but it looks a little funky to me. Are you okay changing it here?

I would return `true` or `false` from this function and then in `main` where you have that `console.error` I would just call the string itself - `console.error("Please enter a valid study from your Firestore database")`

cli.mjs
```diff
@@ -185,19 +242,19 @@ async function deploymentPrompt() {
 }
 
 /** Prompt the user to enter the ID of a study */
-async function studyIDPrompt() {
+// helper to check if the given study (input) is in firestore
+const validateStudyFirebase = async (input) => {
```
RobertGemmaJr: I would keep this as a normal function to keep consistent with the other functions in the file. Generally I only use arrow functions if the function can be defined on a single line, but that's personal preference.
```suggestion
async function validateStudyFirebase(input) {
```

cli.mjs
```diff
@@ -209,14 +266,15 @@ async function studyIDPrompt() {
 }
 
 /** Prompt the user to enter the ID of a participant on the STUDY_ID study */
-async function participantIDPrompt() {
+// helper to check if the given participant (input) is in firestore under study
+const validateParticipantFirebase = async (input) => {
```
RobertGemmaJr:
```suggestion
async function validateParticipantFirebase(input) {
```

RobertGemmaJr (changes requested):
Looking great! There's just a couple small changes as far as the code style

RobertGemmaJr:
Actually @YUUU23 can you test what this PR would look like if merging into `v3.4.2`? All of the commander stuff with the CLI script _should't_ have to wait until v4 to work but I'm not entirely sure.

YUUU23:
> Actually @YUUU23 can you test what this PR would look like if merging into `v3.4.2`? All of the commander stuff with the CLI script _should't_ have to wait until v4 to work but I'm not entirely sure.

It seems like all the forward commits for v4 features would also be on this branch to be merged into 3.4.2. Do you know if it's possible to rebase on a branch with less features that would delete those v4 features 3.4.2 doesn't have? Otherwise, I think I can also move the CLI script code into a new branch based off of 3.4.2 and open a new PR since the CLI code are all in one file.

cli.mjs
YUUU23: This will return from `main()`, so the program would end. 

However, similar to `INVALID_ACTION_ERROR` and `INVALID_DEPLOYMENT_ERROR` (line 19-20), we could potentially create a `INVALID_STUDY_ERROR` / `INVALID_PARTICIPANT_ERROR` to be raised here instead of returning.

RobertGemmaJr: Let's actually call `process.exit(1)` to indicate that the program ended with an error

RobertGemmaJr:
To have this in writing - let's leave it as a change for v4 for right now. Eventually we will release the version and we can include a quick write up for bringing the CLI script into older versions of Honeycomb

cli.mjs
```diff
@@ -18,16 +19,73 @@ let OUTPUT_ROOT; // The root in which data is saved
...
+  .description(`Delete experiment data from Firebase provided study ID and participant ID`)
+  .action((studyID, participantID) => {
+    ACTION = "delete";
+    STUDY_ID = studyID;
+    PARTICIPANT_ID = participantID;
+  });
+commander.parse();
+
+// print message if download or delete provided, along with optional args provided
+if (ACTION != undefined) {
```
RobertGemmaJr:
```suggestion
if (ACTION !== undefined) {
```

cli.mjs
```diff
@@ -18,16 +19,73 @@ let OUTPUT_ROOT; // The root in which data is saved
...
+    `${ACTION} data from Firebase given ${STUDY_ID === undefined ? "" : `study ID: ${STUDY_ID}`} ${PARTICIPANT_ID === undefined ? "" : `and participant ID: ${PARTICIPANT_ID}`}`
+  );
+}
+
 /** -------------------- MAIN -------------------- */
 
 // TODO @brown-ccv #289: Pass CLI arguments with commander (especially for action)
 async function main() {
-  ACTION = await actionPrompt();
+  if (ACTION == undefined) {
```
RobertGemmaJr:
```suggestion
  if (ACTION === undefined) {
```

cli.mjs
```diff
@@ -18,16 +19,73 @@ let OUTPUT_ROOT; // The root in which data is saved
...
+    ACTION = await actionPrompt();
+  }
   DEPLOYMENT = await deploymentPrompt();
   // TODO @brown-ccv #291: Enable downloading all study data at once
-  STUDY_ID = await studyIDPrompt();
+  if (STUDY_ID == undefined) {
```
RobertGemmaJr:
```suggestion
  if (STUDY_ID === undefined) {
```

cli.mjs
```diff
@@ -18,16 +19,73 @@ let OUTPUT_ROOT; // The root in which data is saved
...
+    // when args directly passed in through CLI, check if study is valid
+    const hasStudy = await validateStudyFirebase(STUDY_ID);
+    if (!hasStudy) {
+      console.error("Please enter a valid study from your Firestore database");
+      return;
+    }
+  }
   // TODO @brown-ccv #291: Enable downloading all participant data at once
-  PARTICIPANT_ID = await participantIDPrompt();
+  if (PARTICIPANT_ID == undefined) {
```
RobertGemmaJr:
```suggestion
  if (PARTICIPANT_ID === undefined) {
```

cli.mjs
```diff
@@ -185,22 +242,23 @@ async function deploymentPrompt() {
 }
 
 /** Prompt the user to enter the ID of a study */
+// helper to check if the given study (input) is in firestore
+async function validateStudyFirebase(input) {
+  // subcollection is programmatically generated, if it doesn't exist then input must not be a valid studyID
+  const studyIDCollections = await getStudyRef(input).listCollections();
+  return studyIDCollections.find((c) => c.id === PARTICIPANTS_COL);
+}
```
RobertGemmaJr: Can you move this helper function (along with `validateParticipantFirebase`) into a new section for validation functions? It looks like some of the comments aren't above the correct functions anymore

cli.mjs
```diff
@@ -209,14 +267,15 @@ async function studyIDPrompt() {
 }
 
 /** Prompt the user to enter the ID of a participant on the STUDY_ID study */
+// helper to check if the given participant (input) is in firestore under study
```
RobertGemmaJr: Yeah, it looks like the comment on line 269 is for the function defined on line 277

cli.mjs
```diff
@@ -185,22 +242,23 @@ async function deploymentPrompt() {
...
   return await input({
     message: "Select a study:",
     validate: async (input) => {
       if (!input) return invalidMessage;
-
       switch (DEPLOYMENT) {
         case "firebase":
-          return validateStudyFirebase(input);
+          const res = await validateStudyFirebase(input);
+          return !res ? invalidMessage : true;
```
RobertGemmaJr: We're calling `.find` inside `validateStudyFirebase` so I would rename this to be `studyCollection` or something like that? Because the return of the function is really the Firebase study collection or `undefined`

RobertGemmaJr (changes requested):
Looking good! Just some organizational changes from here!

cli.mjs
```diff
@@ -186,20 +244,14 @@ async function deploymentPrompt() {
...
   return await input({
     message: "Select a study:",
     validate: async (input) => {
       if (!input) return invalidMessage;
-
       switch (DEPLOYMENT) {
         case "firebase":
-          return validateStudyFirebase(input);
+          const studyCollection = await validateStudyFirebase(input);
+          return !studyCollection ? invalidMessage : true;
```
eldu: Okay let's simplify this 😵 I had to bust out a logical operators chart thing. Check my work! 

```suggestion
          return studyCollection || invalidMessage;
```

| SC | !SC | IM | !SC ? IM : 1 | SC or IM | 
| -- | -- | -- | -- | -- |
| 0 | 1 | 0 | 0 | 0 |
| 0 | 1 | 1 | 1 | 1 |
| 1 | 0 | 0 | 1 | 1 |
| 1 | 0 | 1 | 1 | 1 |

cli.mjs
```diff
@@ -225,7 +271,8 @@ async function participantIDPrompt() {
 
       switch (DEPLOYMENT) {
         case "firebase":
-          return validateParticipantFirebase(input);
+          const participantCollection = await validateParticipantFirebase(input);
+          return !participantCollection ? invalidMessage : true;
```
eldu: I think this would be the same as what I had up above
```suggestion
          return participantCollection || invalidMessage;
```

eldu (changes requested):
I commented to simplify some logical expressions. Looks like Rob reviewed this quite a bit ! Small changes, otherwise looks good to go !

YUUU23:
> I commented to simplify some logical expressions. Looks like Rob reviewed this quite a bit ! Small changes, otherwise looks good to go !

Thank you so much for reviewing this!!

YUUU23:
Reverted changes 
Most recent changes in this PR: https://github.com/brown-ccv/honeycomb/pull/531

PR #518, opened 2024-07-30
Ref config env idx

### LINKED ISSUE #363: Re-organize the `config` folder
- config/
    - language.json
    - config.json
    - env.js
    - index.js

`index.js` imports the json files and exports them with the correct name. env.js will do the same but gets re-exported by `index.js`

YUUU23:
Under the `/config` folder: 
- Move environment variables to `env.js`
- Import environment variables into `index.js`, also imports and export other related environment variables such as `SETTINGS` and `LANGUAGES`
- Removed `main.js` file 

- Refactor import statements in trial js and react files to `index.js` instead of `main.js`


src/App/App.jsx
```diff
@@ -5,7 +5,7 @@ import "@fortawesome/fontawesome-free/css/all.css";
 import "./index.css";
 
 // Import configurations and utilities
-import { config, SETTINGS } from "../config/main";
+import { CONFIG, SETTINGS } from "../config/index";
```
RobertGemmaJr: So a very cool thing about the `index.js` file is that it will automatically get resolved in import statements! So you can actually do this for all of them:
```suggestion
import { CONFIG, SETTINGS } from "../config/";
```

src/config/index.js
```diff
@@ -0,0 +1,10 @@
+import packageInfo from "../../package.json";
+import language from "./language.json";
+import settings from "./settings.json";
+import config from "./env.js";
+
+export const taskName = packageInfo.name;
+export const taskVersion = packageInfo.version;
```
RobertGemmaJr: Could you make this UPPER_CASE to match?

src/config/index.js
```diff
@@ -0,0 +1,10 @@
...
+export const LANGUAGE = language;
+export const SETTINGS = settings;
+export const CONFIG = config;
```
RobertGemmaJr: Could you also add the comments and TODOs from `main.js` into this file as well? You can see them if you look at the deleted code in the PR

RobertGemmaJr (changes requested):
Looking good! Just a couple comments and you can remove the `/index` from the import statements! You should be able to do a global find and replace to fix that

RobertGemmaJr:
@YUUU23 Just a heads up that I merged #479 into `feat-v4`. You should be able to remove `taskName` and `taskVersion`. 

They can be used with `import.meta.env.PACKAGE_NAME` and `import.meta.env.PACKAGE_VERSION` but currently we only use them in 1 place - I don't think we need to export a separate variable for them

RobertGemmaJr (approved):
I'm going to approve this because everything looks good! There's just a small merge conflict to fix first and then you can merge this!

PR #519, opened 2024-07-30
Ref event codes json

### LINKED ISSUE #333: Move `event_code`s to `config.json`

YUUU23:
- moved `eventCodes` into its own json file 
- change import/export in `trigger.js` accordingly 
- should we rename the final export in `trigger.js` `EVENT_CODES` instead of `eventCodes`? Also not entirely sure if we want to name this file `config.json` as mentioned in the issue or `eventCodes.json` 

- merged into `ref-config-env-idx` since a lot of the changes there are related to this modification; this branch was also based on `ref-config-env-idx`. This can also be changed to base on `v4` branch if needed.

### DIFF
--- a/src/config/eventCodes.json
+++ b/src/config/eventCodes.json
@@ -1,0 +1,6 @@
+{
+  "fixation": 1,
+  "honeycomb": 2,
+  "open_task": 18,
+  "test_connect": 32
+}
--- a/src/config/trigger.js
+++ b/src/config/trigger.js
@@ -1,3 +1,5 @@
+import event_codes from "./eventCodes.json";
+
 // TODO @brown-ccv #333: Nest this data under "trigger_box" equipment in config.json
 
 // teensyduino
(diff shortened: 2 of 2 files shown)


RobertGemmaJr:
>  This can also be changed to base on `v4` branch if needed.

This will happen automatically! When `ref-config-env-idx` gets merged into `v4` this branch's base will be updated by GitHub

RobertGemmaJr:
Actually, @YUUU23 it looks like this branch should be going into [ref-config-as-ENV](https://github.com/brown-ccv/honeycomb/tree/ref-config-as-ENV)?

YUUU23:
> ref-config-as-ENV

Yes! I will change this.

RobertGemmaJr (approved):
Awesome! I think eventually we can work on moving the `trigger` variable out but let's not worry about that right now. That will hopefully be removed completely in version 5 so we it's not worth it for us to change it right now

PR #520, opened 2024-07-30
ref: rename CONFIG as ENV

### LINKED ISSUE #361: Rename `config` as `ENV`
These are all the variables that are checked in `config/main.js`. Will have to add some of the additional ones well.

YUUU23:
Rename all `CONFIG` as `ENV` 

merged to `ref-config-env-idx` branch as this is the branch that re-organizes `config` files (bigger changes)


src/config/env.js
```diff
@@ -8,7 +8,7 @@ const USE_EEG = import.meta.env.VITE_USE_EEG === "true" && USE_ELECTRON; // Whet
 const USE_PHOTODIODE = import.meta.env.VITE_USE_PHOTODIODE === "true" && USE_ELECTRON; // whether or not the photodiode is in use
 
 // Configuration object for Honeycomb
-const config = {
+const env = {
```
RobertGemmaJr: You can consolidate this a little if you'd like! 
```suggestion
export default {
```

src/config/index.js
```diff
@@ -1,10 +1,10 @@
 import packageInfo from "../../package.json";
 import language from "./language.json";
 import settings from "./settings.json";
-import config from "./env.js";
+import env from "./env.js";
```
RobertGemmaJr: Same here
```suggestion
import ENV from "./env.js";
```

src/config/index.js
```diff
@@ -1,10 +1,10 @@
...
 
 export const taskName = packageInfo.name;
 export const taskVersion = packageInfo.version;
 export const LANGUAGE = language;
 export const SETTINGS = settings;
-export const CONFIG = config;
+export const ENV = env;
```
RobertGemmaJr: Because then you just have to do 
```suggestion
export ENV ;
```

RobertGemmaJr (commented):
Looks amazing!! I left a couple suggestions if you'd like to implement them.

Let me know when you make the changes to [ref-config-env-idx](https://github.com/brown-ccv/honeycomb/tree/ref-config-env-idx) - I would bring those changes into here before merging this branch into that one, if that makes sense

RobertGemmaJr (approved):
Looks great!

PR #521, opened 2024-07-30
ref: change EEG to EVENT_CODES

### LINKED ISSUE #341: Seperate `USE_EEG` config variable into `USE_EVENT_CODES` and `USE_PHOTODIODE`
There should be a config variable for `USE_PHOTODIODE` and `USE_EVENT_MARKER`. `USE_EEG` assumes both - instead instruct the user to turn on both if both are applicable

YUUU23:
Rename `USE_EEG` to  `USE_EVENT_CODES` to separate out `USE_EVENT_CODES` and `USE_PHOTODIODE`

based on `ref-event-codes-json` branch as this is the most recent branch in a series of config folder modifications. Please let me know if I should rebase this!


RobertGemmaJr (changes requested):
Be sure to change the values in `.env.___` too! They're all in the `/env/` folder

env/.env.clinic
```diff
@@ -1,2 +1,2 @@
-VITE_USE_EEG="true"
-VITE_USE_PHOTODIODE="true"
\ No newline at end of file
+USE_EVENT_CODES="true"
```
RobertGemmaJr: We do have to prefix these variables with `VITE_` so they are loaded correctly with the compiler. You can read more here: https://vitejs.dev/guide/env-and-mode#env-files
```suggestion
VITE_USE_EVENT_CODES="true"
```

RobertGemmaJr (changes requested):
Super quick change but otherwise looks great! I love the way you've broken down the issues into super small pull requests

PR #524, opened 2024-07-31
new branch with changes from original callFunction trial able to merg…

YUUU23:
Original code from https://github.com/brown-ccv/honeycomb/pull/515 

I messed up by forgetting that the `callFunction` should be added to 3.4.2 instead of 4.0 (so sorry about this!). This PR carry over those changes from the original PR but based on 3.4.2 instead of 4.0 so it can be merged more easily into this branch without too much changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,18 @@
 {
   "name": "honeycomb",
-  "version": "3.4.0",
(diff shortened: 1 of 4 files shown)


RobertGemmaJr (approved):
Looks great! 🚀

PR #525, opened 2024-08-02
fix: Fix vite build configurations

RobertGemmaJr:
I fixed the configurations here using CJS first. Once this is all set I'll switch back to `type: "module"` 

- Switches from ESM back to CJS
- Adds a `vite.browser.config.js` - the configuration used when using Firebase in the browser
- Adds a baseConfig in `vite.base.config.js` that serves as a base config for BOTH the browser and electron
- The `baseConfig` is merged with the renderer config, not the forge config. This was the fix we needed!

**Misc Changes**
- Little cleanup for eslint with the Firebase and the CLI script


src/App/App.jsx
```diff
@@ -52,6 +52,13 @@ export default function App() {
         "Task Settings": SETTINGS,
       });
 
+      console.log(
+        "ENVIRONMENT",
+        import.meta.env.PACKAGE_NAME,
+        import.meta.env.PACKAGE_VERSION,
+        import.meta.env
+      );
```
RobertGemmaJr, line 61:
```suggestion
     // TEMP: Testing to ensure the config is setup correctly
      console.log(
        "ENVIRONMENT",
        import.meta.env.PACKAGE_NAME,
        import.meta.env.PACKAGE_VERSION,
        import.meta.env
      );
```

RobertGemmaJr:
Here's what I've discovered with the builds themselves:

- `osxSign: {}` is ONLY needed with the universal builds, but will cause the build to crash on platform-specific builds
- The universal build still needs "Open in Rosetta" to be checked off for some reason
- The build for `x64` seems to work fine on an arm mac anyways? Need to test that more
- I also get a "Honeycomb Safe Storage" warning on mac when I go to run the application. I'm not entirely sure where that's coming from

YUUU23 (approved):

PR #526, opened 2024-08-04
Add num blinks

### LINKED ISSUE #354: Event codes should have `name`, `code`, and `numBlinks` properties
These are the three main ways the different trials are differentiated. Would be useful to import from the same place, use as `taskSettings.honeycomb.numBlinks`

YUUU23:
- add `code` and `numBlinks` as nested object to each event code
- modify how even names and code are accessed to reflect the change in object structure 
- change `eventCodes` import in `trigger.js` to be more concise


src/config/eventCodes.json
```diff
@@ -1,6 +1,18 @@
 {
-  "fixation": 1,
-  "honeycomb": 2,
-  "open_task": 18,
-  "test_connect": 32
+  "fixation": {
+    "code": 1,
+    "numbBlinks": 1
```
eldu: typo

```suggestion
    "numBlinks": 1
```

src/config/eventCodes.json
```diff
@@ -1,6 +1,18 @@
...
+  },
+  "honeycomb": {
+    "code": 2,
+    "numBlinks": 1
+  },
+  "open_task": {
+    "code": 18,
+    "numbBlinks": 18
```
eldu: typo

```suggestion
    "numBlinks": 18
```

src/config/eventCodes.json
```diff
@@ -1,6 +1,18 @@
...
+  },
+  "test_connect": {
+    "code": 32,
+    "numbBlinks": 32
```
eldu: typo

```suggestion
    "numBlinks": 32
```

eldu (changes requested):
Looks great! A few typos and some questions here

Where are `numBlinks` being used? It looks like we're just setting it right now

YUUU23:
> Looks great! A few typos and some questions here
> 
> Where are `numBlinks` being used? It looks like we're just setting it right now

I am actually not entirely sure of this; it seems like the most relevant case I found is here: [src/lib/markup/photodiode.js](https://github.com/brown-ccv/honeycomb/blob/922ab5370ef9ea487808d23f076d44a2a8ffc038/src/lib/markup/photodiode.js#L32), we set a `blinkTime`. However, I'm not sure what's the purpose for all eventCodes to have a blink time that is the same as their code.

src/config/eventCodes.json
```diff
@@ -1,6 +1,18 @@
...
-  "honeycomb": 2,
-  "open_task": 18,
-  "test_connect": 32
+  "fixation": {
+    "code": 1,
+    "numBlinks": 1
+  },
+  "honeycomb": {
+    "code": 2,
+    "numBlinks": 1
```
eldu: If the numBlinks should be the same as the code, then this should be `2`?

I feel like this might be a Rob question at this point 😵

eldu (commented):
I think this PR needs @RobertGemmaJr to approve since I don't think I understand the configuration enough to give the okay 😓

YUUU23:
> I think this PR needs @RobertGemmaJr to approve since I don't think I understand the configuration enough to give the okay 😓

Sounds good! I can wait until Rob comes back to discuss more about this. Thank you for reviewing it so far : D !

src/config/eventCodes.json
YUUU23: Yes, this is my mistake without taking a closer look. Let me make this change to make things more uniform and wait for Rob to review the rest.

RobertGemmaJr (commented):
This looks super great! Why don't we move the `code` and `numbBlinks` for each trial type into their trial in the `settings.json` file. This way we can delete the `eventCodes.json` file for some consolidation! All of the values look great!

src/Electron/main.js
```diff
@@ -127,9 +126,9 @@ function handleSetConfig(event, config) {
  * @param {Event} event The Electron renderer event
  * @param {Object} trigger The metadata for the event code trigger
  * @param {string} trigger.comName The COM name of the serial port
- * @param {Object} trigger.eventCodes The list of possible event codes to be triggered
  * @param {string} trigger.productID The name of the product connected to the serial port
  * @param {string} trigger.vendorID The name of the vendor connected to the serial prot
+ * @param {Object} trigger.settings The list of possible event with relative event codes to be triggered
```
YUUU23, line 131: since codes are now embedded under the settings.json, I had to import the entire settings object into the trigger object to access the code

src/config/settings.json
```diff
@@ -16,6 +18,16 @@
         "stimulus": "assets/images/orange.png",
         "correct_response": "j"
       }
-    ]
+    ],
+    "code": 2,
+    "numBlinks": 2
+  },
+  "open_task": {
```
YUUU23, line 25: inserted these 2 events from the original event codes file

src/experiment/procedures/honeycombProcedure.js
```diff
@@ -39,12 +38,13 @@ export const buildHoneycombProcedure = () => {
     choices: honeycombSettings.timeline_variables.map((variable) => variable.correct_response),
     data: {
       // Record the correct_response passed as a timeline variable
-      code: eventCodes.honeycomb,
+      code: honeycombSettings.code,
```
YUUU23, line 41: since `honeycombSettings` is set to `settings.honeycomb`, we can directly access the `code` field

src/experiment/procedures/honeycombProcedure.js
```diff
@@ -39,12 +38,13 @@ export const buildHoneycombProcedure = () => {
...
       correct_response: getJsPsych().timelineVariable("correct_response"),
     },
     on_load: function () {
       // Conditionally flashes the photodiode when the trial first loads
-      if (ENV.USE_PHOTODIODE) pdSpotEncode(eventCodes.honeycomb);
+      if (ENV.USE_PHOTODIODE) pdSpotEncode(honeycombSettings.code);
+      console.log("honeycomb procedure code: " + honeycombSettings.code);
     },
```
RobertGemmaJr: Just removing a console log
```suggestion
      if (ENV.USE_PHOTODIODE) pdSpotEncode(honeycombSettings.code);
    },
```

RobertGemmaJr (approved):
Looks great! I think eventually I will move all of the trigger stuff into the settings file but should be done in another PR. This looks fantastic! 🚀

eldu (approved):

PR #528, opened 2024-08-09
jsPsych as global variable

YUUU23:
- when `jsPsych` instance gets initialized, set global variable `window.jsPsych` to this object 
- remove all function argument that takes in `jsPsych` object
- refactor all usage of `jsPsych` into `window.jsPsych` to be accessed globally 
- add util function to get and return the global `jsPsych` instance
- tested code under different setting (clinic, home, browser, with video)

### DIFF
--- a/src/App/components/JsPsychExperiment.jsx
+++ b/src/App/components/JsPsychExperiment.jsx
@@ -67,6 +67,7 @@ export default function JsPsychExperiment({
    */
   React.useEffect(() => {
     if (jsPsych) {
+      // set jsPsych object as global variable
       window.jsPsych = jsPsych;
       const timeline = buildTimeline(studyID, participantID);
       jsPsych.run(timeline);
--- a/src/App/components/JsPsychExperiment.jsx
+++ b/src/App/components/JsPsychExperiment.jsx
@@ -5,6 +5,7 @@ import React from "react";
 import { ENV } from "../../config/";
 import { buildTimeline, jsPsychOptions } from "../../experiment";
 import { initParticipant } from "../deployments/firebase";
+import { getJsPsych } from "../../lib/utils";
 
 // ID used to identify the DOM element that holds the experiment.
 const EXPERIMENT_ID = "experiment-window";
@@ -70,7 +71,7 @@ export default function JsPsychExperiment({
       // set up jsPsych object as global variable
       window.jsPsych = jsPsych;
       const timeline = buildTimeline(studyID, participantID);
-      window.jsPsych.run(timeline);
+      getJsPsych().run(timeline);
     }
   }, [jsPsych]);
 
--- a/src/App/components/JsPsychExperiment.jsx
+++ b/src/App/components/JsPsychExperiment.jsx
@@ -70,7 +70,7 @@ export default function JsPsychExperiment({
       // set up jsPsych object as global variable
       window.jsPsych = jsPsych;
       const timeline = buildTimeline(studyID, participantID);
-      jsPsych.run(timeline);
+      window.jsPsych.run(timeline);
     }
   }, [jsPsych]);
 
--- a/src/App/components/JsPsychExperiment.jsx
+++ b/src/App/components/JsPsychExperiment.jsx
@@ -67,7 +67,7 @@ export default function JsPsychExperiment({
    */
   React.useEffect(() => {
     if (jsPsych) {
-      // set jsPsych object as global variable
+      // set up jsPsych object as global variable
       window.jsPsych = jsPsych;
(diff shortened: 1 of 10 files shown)


src/App/components/JsPsychExperiment.jsx
```diff
@@ -51,7 +51,7 @@ export default function JsPsychExperiment({
       tempJsPsych.data.addProperties({
         app_name: import.meta.env.PACKAGE_NAME,
         app_version: import.meta.env.PACKAGE_VERSION,
-        app_commit: await window.electronAPI.getCommit(),
+        // app_commit: await window.electronAPI.getCommit(),
```
YUUU23: commented this out because I wasn't sure why it errored with reading as undefined when running in browser

eldu: @YUUU23 What was the error you were getting?

eldu (changes requested):
I see that you've created a `getJsPsych` utility function, but are using `window.jsPsych` in all of the refactored lines. 

What reasoning is behind this? From my perspective it seems like you should use the utility function throughout.

YUUU23:
> I see that you've created a `getJsPsych` utility function, but are using `window.jsPsych` in all of the refactored lines.
> 
> What reasoning is behind this? From my perspective it seems like you should use the utility function throughout.

I think when I talked with Rob about this, he noted that maybe first we use the `window.jsPsych` method for a draft PR to ensure that everything works, and include the utilities function so that it would be more straightforward for developers with not much programming experiences to be able to access `jsPsych` through a utility function instead of accessing the global variable which is not as straightforward. 

I do agree maybe accessing through the utility function throughout Honeycomb may be more consistent and flexible to modifications. The only thing I am unsure about is if it's necessary since accessing this object is only a oneliner. 

Please let me know if you have any further thoughts on this! I'm leaning towards refactoring to access through util function throughout.

src/App/components/JsPsychExperiment.jsx
YUUU23: Just re-ran honeycomb with different settings and it seems like this error went away! Not sure if it was a previous dependency issue or something but seems like everything is ok now!

Uncommented this line in the new commit also.

eldu:
> > I see that you've created a `getJsPsych` utility function, but are using `window.jsPsych` in all of the refactored lines.
> > What reasoning is behind this? From my perspective it seems like you should use the utility function throughout.
> 
> I think when I talked with Rob about this, he noted that maybe first we use the `window.jsPsych` method for a draft PR to ensure that everything works, and include the utilities function so that it would be more straightforward for developers with not much programming experiences to be able to access `jsPsych` through a utility function instead of accessing the global variable which is not as straightforward.
> 
> I do agree maybe accessing through the utility function throughout Honeycomb may be more consistent and flexible to modifications. The only thing I am unsure about is if it's necessary since accessing this object is only a oneliner.
> 
> Please let me know if you have any further thoughts on this! I'm leaning towards refactoring to access through util function throughout.

Let's use the utility function that we'd be directing folks to use. It might unclear to folks if they see `window.jsPsych` used everywhere instead of the util function. And it would be good to test to make sure the utils function works (i believe in it fully honestly).

 So take out the `window.jsPsych` and let's use the utility function throughout!

eldu (changes requested):
I added a comment to the other thread. Basically, let's use the utility function throughout instead of `window.jsPsych` !

Thanks for the great work Megan !

YUUU23:
> > > I see that you've created a `getJsPsych` utility function, but are using `window.jsPsych` in all of the refactored lines.
> > > What reasoning is behind this? From my perspective it seems like you should use the utility function throughout.
> > 
> > 
> > I think when I talked with Rob about this, he noted that maybe first we use the `window.jsPsych` method for a draft PR to ensure that everything works, and include the utilities function so that it would be more straightforward for developers with not much programming experiences to be able to access `jsPsych` through a utility function instead of accessing the global variable which is not as straightforward.
> > I do agree maybe accessing through the utility function throughout Honeycomb may be more consistent and flexible to modifications. The only thing I am unsure about is if it's necessary since accessing this object is only a oneliner.
> > Please let me know if you have any further thoughts on this! I'm leaning towards refactoring to access through util function throughout.
> 
> Let's use the utility function that we'd be directing folks to use. It might unclear to folks if they see `window.jsPsych` used everywhere instead of the util function. And it would be good to test to make sure the utils function works (i believe in it fully honestly).
> 
> So take out the `window.jsPsych` and let's use the utility function throughout!

Just refactored this in the new commit!

eldu (approved):
Amazing work Megan !

PR #529, opened 2024-08-09
ref: change all func taking in jspsych to const syntax / trials to co…

YUUU23:
After removing`jsPsych` passed in as params in `jsPsych-global` branch: 
- For `procedures`: change to `const` function syntax, call functions in original place  
- For `trials`: change to `const` jsPsych trial object syntax, move variable declarations outside the object declaration when necessary 
- javadoc: remove tags requiring `jsPsych` as a parameter 
- tested with running browser, dev home, dev clinic, dev clinic with video

### DIFF
--- a/src/experiment/trials/camera.js
+++ b/src/experiment/trials/camera.js
@@ -9,111 +9,105 @@ const WEBCAM_ID = "webcam";
 
 /**
  * A trial that begins recording the participant using their computer's default camera
- * @param {Object} jsPsych The jsPsych instance being used to run the task
+ *
  * @returns {Object} A jsPsych trial object
  */
 // TODO @brown-ccv #301: Use jsPsych extension, deprecate this function
 // TODO @brown-ccv #343: We should be able to make this work on both electron and browser?
 // TODO @brown-ccv #301: Rolling save to the deployment (webm is a subset of mkv)
-export function buildCameraStartTrial() {
-  return {
-    timeline: [
-      {
-        // Prompts user permission for camera device
-        type: initializeCamera,
-        include_audio: true,
-        mime_type: "video/webm",
+export const buildCameraStartTrial = {
+  timeline: [
+    {
+      // Prompts user permission for camera device
+      type: initializeCamera,
+      include_audio: true,
+      mime_type: "video/webm",
+    },
+    {
+      // Helps participant center themselves inside the camera
+      type: htmlButtonResponse,
+      stimulus: function () {
+        const videoMarkup = tag("video", "", {
+          id: WEBCAM_ID,
+          width: 640,
+          height: 480,
+          autoplay: true,
+        });
+        const cameraStartMarkup = p(LANGUAGE.trials.camera.start);
+        const trialMarkup = div(cameraStartMarkup + videoMarkup, {
+          class: "align-items-center-col",
+        });
+        return div(trialMarkup);
       },
-      {
-        // Helps participant center themselves inside the camera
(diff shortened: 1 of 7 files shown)


src/experiment/trials/camera.js
```diff
@@ -9,111 +9,105 @@ const WEBCAM_ID = "webcam";
...
-  };
-}
+    },
+  ],
+};
 
 /**
  * A trial that finishes recording the participant using their computer's default camera
  *
- * @param {Number} duration How long to show the trial for
```
YUUU23, line 94: removed this since it looks like we don't use this param anymore too

src/experiment/trials/camera.js
```diff
@@ -9,111 +9,105 @@ const WEBCAM_ID = "webcam";
 
 /**
  * A trial that begins recording the participant using their computer's default camera
- * @param {Object} jsPsych The jsPsych instance being used to run the task
+ *
  * @returns {Object} A jsPsych trial object
  */
 // TODO @brown-ccv #301: Use jsPsych extension, deprecate this function
```
eldu: Since this is no longer a function, more docstring updates. Also, I would have suggested renaming the variable, but since it looks like it's going to be deprecated, I think it's fine to leave as is

```suggestion
 *
 * @type {Object} A jsPsych trial object
 */
// TODO @brown-ccv #301: Use jsPsych extension, deprecate this variable
```

src/experiment/trials/camera.js
```diff
@@ -9,111 +9,105 @@ const WEBCAM_ID = "webcam";
...
-  const recordingEndMarkup = h1(LANGUAGE.trials.camera.end);
-
-  return {
-    type: htmlKeyboardResponse,
-    stimulus: div(recordingEndMarkup),
-    trial_duration: 5000,
-    on_start: function () {
-      // Complete the camera recording
+const recordingEndMarkup = h1(LANGUAGE.trials.camera.end);
+export const buildCameraEndTrial = {
```
eldu: The docstring above doesn't seem to match what it's supposed to be for. The docstring is just above `recordingEndMarkup` rather than `buildCameraEndTrial`

src/experiment/trials/camera.js
```diff
@@ -9,111 +9,105 @@ const WEBCAM_ID = "webcam";
...
+    if (!ENV.USE_ELECTRON) {
+      throw new Error("video recording is only available when running inside Electron");
+    }
 
-      const cameraRecorder = window.jsPsych.pluginAPI.getCameraRecorder();
-      if (!cameraRecorder) {
-        console.error("Camera is not initialized, no data will be recorded.");
-        return;
-      }
+    const cameraRecorder = window.jsPsych.pluginAPI.getCameraRecorder();
```
eldu: I commented on the other out `window.jsPsych`. What's the order of these PRs?

eldu (changes requested):
Looks good! Great job Megan! A few questions and changes

src/experiment/trials/camera.js
YUUU23: Please let me know if this is not what you mean by order, but I think we should merge this branch into `jsPsych-global` and then merge `jsPsych-global` into `feat-v4`

eldu (approved):
Okay! This one looks good. I did request some changes to https://github.com/brown-ccv/honeycomb/pull/528#pullrequestreview-2235734231 which affects this PR in some ways. 

So resolve those and this one looks good to go!

PR #552, opened 2025-08-15
path args to SerialPort needs to be an object after v10

### LINKED ISSUE #551: [BUG] Calling `new SerialPort(path)` throws "path is not defined" instead of opening port`
**Title:** [Docs] Clarify that `new SerialPort()` requires an options object in v10+  

TODO:
- Fix serialPort.js to resolve the following bug. 

**Description:**  
In older versions of serialport (pre-v10), the constructor supported passing a string path as the first argument. In serialport v10+ (tested on v12.0.0), the constructor now **requires** an options object with a `path` property. Passing a string path (even if defined in a variable) now throws an error.

Older code example:
```js
const port = new SerialPort('COM4');
```

**Steps to Reproduce:**
1. Install serialport v12.0.0.
2. Run the reproduction code below.
3. Observe the error output.

```js
const { SerialPort } = require('serialport');
const path = 'COM4';
const port = new SerialPort(path);
```

From `SerialPortStream`:
```
TypeError: path is not defined
```

**Expected Behavior:**  
Docs should explicitly state that the only supported constructor form in v10+ is the options object form with a `path` property.

Current Interface from docs: https://serialport.io/docs/guide-usage

```js
const port = new SerialPort({
  path: 'COM4',
  baudRate: 9600
});
```

**Actual Behavior:**  
Older code/examples using the string path shorthand no longer work in v10+.

**Environment:**
- serialport: 12.0.0
- Node.js: vXX.XX.X
- Platform: Windows 10 x64
- Electron: 34.5.1 (reproducible in plain Node too)

**Suggested Fix:**  
Update the API docs to clarify:
- Only the options object form is supported in v10+.
- The string path constructor overload was removed after v9.

fordmcdonald:
### Interface to Create SerialPort is `new SerialPort({path: path})` 

I also resolved an instance of using the deprecated `port.comName` to `port.path`.
 
See Docs: https://serialport.io/docs/guide-usage
Closes #551

### DIFF
--- a/public/electron/serialPort.js
+++ b/public/electron/serialPort.js
@@ -15,7 +15,7 @@ function getDevice(portList, comVendorName, productId) {
     const comName = comVendorName;
     return portList.filter(
       // Find the device with the matching comName
-      (device) => device.comName === comName.toUpperCase() || device.comName === comName
+      (device) => device.path === comName.toUpperCase() || device.path === comName
     );
   } else {
     const vendorId = comVendorName;
@@ -47,7 +47,7 @@ async function getPort(comVendorName, productId) {
   const device = getDevice(portList, comVendorName, productId);
   try {
     const path = device[0].path;
-    const port = new SerialPort(path);
+    const port = new SerialPort({path: path});
     return port;
   } catch {
     return false;


RobertGemmaJr (approved):
LGTM 🚀

